PR #4, opened 2016-12-16
Fixed yaml line terminators in a hand edited file

mkanoor:
The YAML line terminators are different in hand edited and programmatically generated YAML files.
Changed to use the programmatically generated YAML file


Fryguy:
@mkanoor I'm working on a spec that can autodetect this situation at pull request time, so it will fail on Travis.

PR #13, opened 2016-12-20
Load manageiq-content tasks before app Rakefile

bdunne:
Prevent manageiq-content tasks from inheriting the `app:` prefix


bdunne:
@Fryguy Please review

PR #14, opened 2016-12-21
Leverage ManageIQ::Environment

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/13299
Less duplication of setup code, single place to keep up to date.


bin/setup
```diff
@@ -1,28 +1,18 @@
 #!/usr/bin/env ruby
 require 'pathname'
 
-GEM_ROOT = Pathname.new File.expand_path('../../',  __FILE__)
+GEM_ROOT = Pathname.new(__dir__).join("..")
 
 Dir.chdir(GEM_ROOT) do
   if File.exist?("spec/manageiq")
```
Fryguy: @bdunne Can you also change this to not do if File.exist? but instead File.file?.  This is because if I have it symlinked I probably *don't* want it to update at all.

bin/setup
```diff
@@ -1,28 +1,18 @@
...
     puts "== Updating manageiq sample app =="
     system "cd spec/manageiq && git pull"
   else
     puts "== Cloning manageiq sample app =="
-    system "git clone git@github.com:ManageIQ/manageiq.git --depth 1 spec/manageiq"
+    source = ENV["CI"] ? "https://github.com/ManageIQ/manageiq.git" : "git@github.com:ManageIQ/manageiq.git"
```
Fryguy: That's weird...why is this needed?

bdunne: I want to share the same script between users and Travis.  Travis doesn't have any ssh keys, so it can't use the SSH option.

durandom:
I guess  this is needed to make master 🍏 again?

At least amazon [fails](https://travis-ci.org/ManageIQ/manageiq-providers-amazon/builds/188049580) currently because of adding `./tools/ci/setup_ruby_environment.rb` [here](https://github.com/ManageIQ/manageiq/pull/13278/files#diff-b415c2cf31c9d03a2de7e17f7a7fdd30R1) which assumes the script is run from the root of travis CI - which is not the case, at least for the provider repos.

durandom:
Also, I foresee more breaking changes down the road that affect all plugins, providers, ui's etc.

How about refactoring the provider generator into a plugin generator and a provider generator (that uses the plugin generator).
Then use the generator to propagate changes across the plugin repos.

bin/update
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env ruby
+require 'pathname'
+
+gem_root = Pathname.new(__dir__).join("..")
+
+unless gem_root.join("spec/manageiq").symlink?
+  puts "== Updating manageiq sample app =="
+  system("git pull", :chdir => gem_root.join("spec/manageiq"))
+end
```
durandom: maybe add a notification that you are not updating, because its symlinked?

```
else
  puts "== SKIP updating of manageiq sample app because its symlinked =="
end
```

I have the feeling that will save some hours of guessing multiplied by all people that use the plugins

bin/setup
```diff
@@ -1,28 +1,12 @@
...
-    system "cp spec/manageiq/config/database.pg.yml spec/manageiq/config/database.yml"
-  end
-
-  unless File.exist?("spec/manageiq/certs/v2_key")
-    puts "\n== Copying sample certs/v2_key =="
-    system "cp spec/manageiq/certs/v2_key.dev spec/manageiq/certs/v2_key"
-  end
+unless gem_root.join("spec/manageiq").exist?
+  puts "== Cloning manageiq sample app =="
+  system "git clone https://github.com/ManageIQ/manageiq.git --depth 1 spec/manageiq"
```
durandom: just thinking aloud: would it make sense to symlink on certain conditions?

Like, I have all plugins in `manageiq/plugins` checked out, and if `gem_root.join('../../lib/vmdb')` exists, I probably want to symlink...

bdunne: Our assumption (@Fryguy & me) is that `bin/setup` should only be run for setup, and would provide a default setup (and skip the clone if already symlinked).  Anyone interested in symlinking would know how to do so and would do that manually.

bin/update
Fryguy: I like this idea...better to be explicit to the user so there's no guessing.

PR #26, opened 2017-01-24
Refresh based on all saved Amazon EmsEvents

Ladas:
Refresh based on all saved Amazon EmsEvents in a way that
we can override the behaviour for single EventTypes
from automate.

https://github.com/ManageIQ/manageiq-providers-amazon/pull/93
needs to be reverted, since it changed all unknown event_types
to one. And The list of recognized events was in the code.

THis PR has the same behaviour, but all is configurable from
Automate.

fixes:
https://bugzilla.redhat.com/show_bug.cgi?id=1371222
https://github.com/ManageIQ/manageiq/issues/10591

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/_missing.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/_missing.yaml
@@ -7,4 +7,6 @@ object:
     name: ".missing"
     inherits: 
     description: 
-  fields: []
+  fields:
+  - rel4:
+      value: "/System/event_handlers/event_action_refresh?target=ems"


gmcculloug:
@Ladas This feels like a drastic solution to the issue of identifying new resources, like IPs and security groups, from Amazon.

This is saying that every event we get from Amazon, which does not already have a defined instance in automate, will trigger a provider refresh.  We may not have a lot of events from them now but that could change in the future.

We should probably be doing something closer to OpenStack and defining the events we really want to react to.  (See https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/EmsEvent/OpenStack.class)

@blomquisg Thoughts?

Ladas:
@gmcculloug well, we already have it being refreshed on each event in https://github.com/ManageIQ/manageiq-providers-amazon/pull/93 So this should be just more configurable way. 

I was thinking about defining of each event, but with AWS Config, CloudWatch and CloudTrail, we would need to define hundreds of events, that we actually want to cause refresh. So maintaining that high number of events is a bit error prone I think.

So this seems to me like a nice default, that can be easily disabled? Plus the events subscribed are changeable on the AWS side, so you can set it to send only few events.

blomquisg:
I think that `missing` is going to catch too many event types.  At best, I think this should be in the documentation on how to enable refresh for all events...instead of being part of the base install.

Ladas:
@blomquisg @gmcculloug right, but that is exactly what we want. Refresh based on any event that happened. :-) Also that is what the BZ ^ wants.

Important to say here is, that I don't do traditional targeted refresh here. The new targeted refresh collects all queued targets for refresh together (where the EmsEvent is a new type of target)

From those targets, it extrapolates ems_refs of any changed element (vm, network_port, subnet, network, etc. It's like 16 types of targets now, will be more).

Then using a set of all things that changed, we do 1 refresh, that refreshes all the changed things.

So, it's crucial to track every little change. :-) But even with targeted refresh disabled, we still want to do refresh when anything has changed, so as the BZ says.

blomquisg:
>  but that is exactly what we want. Refresh based on any event that happened

That's the part I don't agree with.  

Take this even on RHEVM for instance:  https://github.com/ManageIQ/manageiq-content/blob/3a268de8240d561e0400c35ba9de2f91d6dc6e5a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/user_logged_in_vm.yaml

In this case, we have a mapped action for it, which isn't refresh.  However, it's possible that there are many other events that RHEVM emits that should not result in a refresh either because of their frequency of occurrence or because it represents no valuable inventory state change.

The fact that we got an event from a provider doesn't *always* mean that there was an interesting inventory change.  It's probably true that *most* events represent inventory changes.  But, I don't think we should make users tune filtered events because we treat all non-mapped events as an inventory state change.  And, as far as I can see, the only way to deal with non-mapped events that shouldn't trigger inventory refreshes is to filter out them completely.

Ladas:
@blomquisg @gmcculloug Right, well for RHEVM, things might be different. But this changes method_missing only for AWS. An alternative is indeed write down the hundreds of events in automate, but I don't really want to do that and I don't really want to manage the changing list of events. Especially when they didn't bother to document in one place, what are all the events. :-)

So for AWS, it should be ok to refresh any event in method missing. You are always free to define the event the old way, if you need a different behavior. Or even change the method missing, you should be able to do a conditional action by looking at the name of the event.

Then I will just define like 30 interesting events into the timelines.

Then all is great with little changes. :-)

In aws, it event refers e.g. 'instanceId', I don't care what is does, but that Instance will be refreshed. Then 1 line in automate and 1 line in parser handles targeted refresh for about 40 Events. Same for 'volumeId', 'subnetId', etc.

We might cause refresh for data changes we do not track, but we will not miss anything, which seems more important. Also, if you subscribe events about e.g. Users, those will be sent for refresh, but will not refresh anything, since the references are unknown (until we say, userId is connected to a refreshable object)

Ladas:
@gmcculloug @blomquisg https://github.com/ManageIQ/manageiq-providers-amazon/pull/114 was merged. So BZ referenced is no longer fixed. We need this PR to fix it.

In AWS you pay for every event and we work on automatic configuration of events we will receive. https://github.com/ManageIQ/manageiq-providers-amazon/issues/126

That means, we should receive only events we are interested in. Then refreshing on any event is desirable as a default behavior.

gmcculloug:
Automate is just a pass-through to allow customization of event processing.  I am not close enough to the provider details to say how much of an impact this will really have.  If we know the events will be limited due to having to configure/pay for each one this may be ok.

I would caution against statements like:

> Also that is what the BZ ^ wants.

They should be stating the symptoms of a problem more then defining the solution.

@blomquisg I leave this one to you.  In the end if we are receiving too many refresh requests we may need to come up with a completely different solution.  At the moment there are only two choices I see: 1) Individually define each event.  2) Processing all events.

Ladas:
@gmcculloug is there a possibility to use regexp or wilcards? I would like to e.g. run refresh for all events starting with AWS_ALARM_* OR would it be possible to somehow code this into the method missing?

durandom:
https://bugzilla.redhat.com/show_bug.cgi?id=1371222 needs this to do a refresh on AWS_EC2_EIP_CREATE events.

TBH I think queuing a full refresh on every event also too hard.
I would maintain a base list of events for full refresh (or targeted soon) and add AWS_EC2_EIP_CREATE to that list.
That list involves those automate yml files I guess?

> At best, I think this should be in the documentation on how to enable refresh for all events...

Dont we already have docs for this? Sound like a very common customization use case to me.

kbrock:
Is there a mechanism so we can detect new event types and sanity check it later?

durandom:
> Is there a mechanism so we can detect new event types and sanity check it later?

Not that I know of. It's not even in the UI to show events that are _not_ in settings.yml. So, I believe, the only way to see all events that are connected to an object is actually to look into the DB 😑

blomquisg:
Some of the concern here is that this would have every event from Amazon causing a full refresh of the EMS.

However, now we have the capability to do targeted graph-based refreshes in AWS.  So, we might want to revisit this idea.  Maybe there should be a way to indicate what type of thing should be updated in the event handler?  Then, this could be setup to do a targeted refresh based on the event.

Eventually maybe this could be changed into the streaming event -> inventory updates that @agrare is playing with for VMWare.

Ladas:
Targeted refresh based on any event here https://github.com/ManageIQ/manageiq-content/pull/180/files

Ladas:
replaced by https://github.com/ManageIQ/manageiq-content/pull/180/files

PR #27, opened 2017-01-25
Move test dependencies to the gemspec

bdunne:


bdunne:
@Fryguy Please review

PR #29, opened 2017-01-30
Don't wait for refresh if stack no longer exists

bzwei:
https://bugzilla.redhat.com/show_bug.cgi?id=1414989

We don't have a good way to determine whether a refresh completes. Currently we are relying on whether a target stack appears in VMDB. The reported problem occurs when a stack is failed (rolled back as reported in the BZ), it no longer exists in the provider and nor can it exist in VMDB. In this case we should not retry and wait for refresh to complete. 

@mkanoor @gmcculloug please review

### DIFF
--- a/spec/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned_spec.rb
+++ b/spec/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned_spec.rb
@@ -110,6 +110,16 @@ describe ManageIQ::Automate::Cloud::Orchestration::Provisioning::StateMachines::
       expect(ae_service.root['ae_result']).to eq("retry")
     end
 
+    it "does not need to wait for refresh completes if the stack has been removed from provider" do
+      ae_service.set_state_var('provider_last_refresh', true)
+      ae_service.set_state_var('deploy_result', 'error')
+      allow(svc_model_service)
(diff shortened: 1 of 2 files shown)


spec/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned_spec.rb
```diff
@@ -110,6 +110,16 @@
       expect(ae_service.root['ae_result']).to eq("retry")
     end
 
+    it "does not need to wait for refresh completes if the stack has been removed from provider" do
+      ae_service.set_state_var('provider_last_refresh', true)
+      ae_service.set_state_var('deploy_result', 'error')
+      allow(svc_model_service)
+        .to receive(:orchestration_stack_status) { ['check_status_failed', 'stack does not exist'] }
+      amazon_stack.update_attribute(:ems_ref, nil)
```
gmcculloug: @bzwei Per robocop warning would be better to use `amazon_stack.update_attributes(:ems_ref => nil)`

simaishi:
@bzwei This doesn't cherry-pick cleanly. Please create an Euwe specific PR.

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/14150

simaishi:
Backported to Darga via https://github.com/ManageIQ/manageiq/pull/14293

PR #33, opened 2017-02-03
Automate - Copy ConfigurationManagement to AutomationManagement.

billfitzgerald0120:
Copied entire ConfigurationManagement to new AutomationManager namespace.


content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb
```diff
@@ -0,0 +1,121 @@
+#
+# Description: Launch a Ansible Job Template and save the job id
+#              in the state variables so we can use it when we
+#              wait for the job to finish.
+#
+
+class LaunchAnsibleJob
```
mkanoor: @billfitzgerald0120 
Since this is a new method should we add the module around the class

PR #42, opened 2017-02-09
instance_of check doesn't work with Drb Reference object

mkanoor:
https://github.com/ManageIQ/manageiq/issues/13844

When we test Automate methods we went a little too far to compare the passed in service using instance_of, it only works in tests but not in real use when we have DRb running the methods.
The DRb reference object fails the instance_of check.

Removed a test and code that validates service using instance_of

@pkomanek @Ladas @bzwei


Ladas:
@mkanoor cool, looks like my local fix :+1:

gmcculloug:
Fixes https://github.com/ManageIQ/manageiq/issues/13844

PR #43, opened 2017-02-09
Cause the clone task to restart if CheckProvision is in error

mansam:
Cause the Clone from VM task to restart from the Placement step if CheckProvision enters an error state.

@gmcculloug this would go along with https://github.com/ManageIQ/manageiq/pull/13608 to ensure that the exception that can be raised during cloning results in the statemachine restarting.


gmcculloug:
I can see something like this method existing as an example but it cannot be wired into the state machine by default.  As written this would put the provision into a loop.  It also does not take into account manual placement which means we would not change the settings but continually retry the provision.

mansam:
@gmcculloug Is there a way I should be writing this to manually take advantage of the max_retries setting, or is there some other way I should be approaching this?

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_placement.rb
```diff
@@ -0,0 +1,29 @@
...
+# from the Placement step.
+# Required inputs: status
+#
+
+prov = $evm.root['miq_provision']
+unless prov
+  $evm.log(:error, "miq_provision object not provided")
+  exit(MIQ_STOP)
+end
+status = $evm.inputs['status']
```
mkanoor: @mansam 
You would need some condition here to check that Placement really failed, if you had a script error this would keep running for ever.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_placement.rb
```diff
@@ -0,0 +1,29 @@
...
+# Update Status Message
+updated_message  = "[#{$evm.root['miq_server'].name}] "
+updated_message += "VM [#{prov.get_option(:vm_target_name)}] "
+updated_message += "Step [#{$evm.root['ae_state']}] "
+updated_message += "Status [#{status}] "
+updated_message += "Message [#{prov.message}] "
+prov.miq_request.user_message = updated_message
+prov.message = status
+
+if $evm.root['ae_result'] == "error"
```
mkanoor: @mansam 
In line 13 we set ae_result to 'restart' so this line will not get executed unless you are going to put a condition around setting the ae_result to restart

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_placement.rb
mansam: @mkanoor Is the exception that resulted in the error state exposed to automate so that the script can make a decision based on it, or do I have to determine another way? Additionally, is there a way to have this automatically respect `max_retries`, or am I going to have to implement my own mechanism?

mkanoor: @mansam 
@gmcculloug Correct me if I am wrong.
There are 2 state machines here, 1 internal inside the app/models and the other one in Automate.

Internal State Machine:
The start_clone_task is an asynchronous process that talks to the provider to get the provisioning started.
Then the poll_clone_complete waits with a retry and then signals the vmdb in database after it has done an ems_refresh.

https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/provision/state_machine.rb#L28



The external state machine define in Automate is in the check_provisioned loop waiting for the machine to show up  in our database my checking the status of the task.

https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb#L7

So I am guessing we would have to depend on the provisioning task to provide more errors from the cloning process, I don't think we have an explicit column that stores the error code from the provider. So I think we would have to parse the tasks error message and see if we can glean something from there to trigger a retry on placement.

mansam: @mkanoor Parsing the error message seems like a viable solution for this. Any thoughts on how I can make this work with the max_retries mechanism, or if I should even try?

gmcculloug: Max retry is a per-step check so it will not work in this case.  You would need to use `set_state_var` and `get_state_var` to monitor the number of attempts.  https://github.com/pemcg/manageiq-automation-howto-guide/blob/master/chapter12/state_machines.md#saving-variables-between-state-retries

mansam: @gmcculloug Thanks for the clarification.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_placement.rb
```diff
@@ -9,12 +9,25 @@
...
   exit(MIQ_STOP)
 end
-status = $evm.inputs['status']
-$evm.root['ae_result'] = 'restart' 
-$evm.root['ae_next_state'] = 'Placement'
-$evm.log("info", "Provisioning #{prov.get_option(:vm_target_name)} failed, retrying placement.")
+
+DEFAULT_RETRIES = 3
+max_placement_retries = $evm.inputs['retries'] || DEFAULT_RETRIES
+if prov.message.include?("An error occurred while provisioning Instance")
```
mkanoor: @mansam 
Is this a good enough message to capture all cloud placement errors.
Do you want to expose the "placement error message" as a passed in value from $evm.inputs so someone can customize it if they want
```
DEFAULT_PLACEMENT_ERROR_MESSAGE = "An error occurred while provisioning Instance"
placement_error_message = $evm.inputs['placement_error_message'] || DEFAULT_PLACEMENT_ERROR_MESSAGE
```

mansam: I'm only aware of the one exception, but I think exposing the error message seems like a good way to go about things.

mansam:
@mkanoor I think this is ready for you to rereview. :)

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
```diff
@@ -0,0 +1,69 @@
...
+                @prov.miq_request.user_message = updated_message
+                @prov.message = status
+              end
+
+              def retry_state
+                if @prov.message.include?(@error_to_catch) && (@retry_number < @max_retries)
+                  @handle.root['ae_result'] = 'restart'
+                  @handle.root['ae_next_state'] = @restart_from_state
+                  @handle.log("info", "Provisioning #{@prov.get_option(:vm_target_name)} failed, retrying #{@restart_from_state}.")
+                  @handle.set_state_var(:state_retries, @retry_number + 1)
```
mkanoor: @mansam 
Would placement_retries be more appropriate here instead of state_retries.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
```diff
@@ -0,0 +1,69 @@
...
+
+              def update_status_message
+                status = @handle.inputs['status']
+                updated_message  = "[#{@handle.root['miq_server'].name}] "
+                updated_message += "VM [#{@prov.get_option(:vm_target_name)}] "
+                updated_message += "Step [#{@handle.root['ae_state']}] "
+                updated_message += "Status [#{status}] "
+                updated_message += "Message [#{@prov.message}] "
+                @prov.miq_request.user_message = updated_message
+                @prov.message = status
```
mkanoor: @mansam 
Should the message indicate that we are retrying placement

spec/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        ro["miq_provision"] = provision
+        ro["miq_server"]    = miq_server
+        ro["ae_result"]     = "error"
+        ro["ae_next_state"] = "some_state"
+      end
+    end
+    let(:ae_service) { Spec::Support::MiqAeMockService.new(root_object) }
+
+    it "retries and increments count because of a matching error" do
+      allow(ae_service).to receive(:inputs).and_return({})
```
mkanoor: @mansam @gmcculloug 
If this PR https://github.com/ManageIQ/manageiq/pull/13912 gets merged then you dont need to have this line.
You can set input to be a hash from the spec if you need it.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
mansam: No, the entire script has been refactored to be more generic since there was no reason it needed to be placement-specific.

spec/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state_spec.rb
mansam: Okay great

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
mansam: Yeah, probably a good idea to indicate in the status message what is happening in the event that the state is retried.

mansam:
@mkanoor any other changes that need to be made?

mansam:
Hi @mkanoor, anything else you need me to do on this?

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/clone_to_vm.yaml
```diff
@@ -10,6 +10,8 @@ object:
   fields:
   - PreProvision:
       value: "/Cloud/VM/Provisioning/StateMachines/Methods/PreProvision_Clone_to_VM#${/#miq_provision.source.vendor}"
+  - CheckProvisioned:
+      on_error: retry_state(status => 'Error Creating VM')
```
gmcculloug: Same change would be required to the `template.yaml` in the same directory.  One instance is used when you select the template from `Lifecycle -> Provision Instances` and the other when you select the image and use `Lifecycle -> Provision Instances using this image`

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
```diff
@@ -0,0 +1,72 @@
...
+# Required inputs: status
+# Optional inputs: max_retries, error_to_catch, restart_from_state
+#
+module ManageIQ
+  module Automate
+    module Service
+      module Generic
+        module StateMachines
+          module VMProvision_VM
+            class RetryState
```
mansam: Noted, thanks

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
```diff
@@ -0,0 +1,72 @@
...
+                updated_message += "Message [#{@prov.message}] "
+                if @handle.root['ae_result'] == 'restart'
+                  updated_message += "Restarting from #{@restart_from_state} step"
+                end
+                @prov.miq_request.user_message = updated_message
+                @prov.message = status
+              end
+
+              def retry_state
+                if @prov.message.include?(@error_to_catch) && (@retry_number < @max_retries)
```
mansam: I'm definitely open to suggestions on a better way to do this. Parsing the exception is what was previously recommended to me, but it's definitely fragile.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/retry_state.rb
```diff
@@ -0,0 +1,72 @@
...
+                @retry_number = if @handle.state_var_exist?(:state_retries)
+                                  @handle.get_state_var(:state_retries)
+                                else
+                                  0
+                                end
+              end
+
+              def main
+                @handle.log("info", "Starting retry_state")
+                retry_state
```
mansam: My intention is just to affect Openstack provisioning. I might have made a mistake with my scoping.

PR #44, opened 2017-02-09
Automate - Service/Generic postprocess test.

billfitzgerald0120:
Test for new Service PostProcess StateMachine method.

Added begin and rescue to method.


billfitzgerald0120:
@mkanoor 
Please review

PR #45, opened 2017-02-09
Automate - Service/Generic refresh test.

billfitzgerald0120:
Test for new Service Refresh StateMachine method.

Added begin and rescue section to method.

### DIFF
--- a/content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/refresh.rb
+++ b/content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/refresh.rb
@@ -1,6 +1,7 @@
 #
 # Description: This class calls a refresh
 #
+
 module ManageIQ
   module Automate
     module Service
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/refresh.rb
```diff
@@ -13,9 +14,18 @@ def initialize(handle = $evm)
               end
 
               def main
-                @handle.log("info", "Starting Refresh")
-                service.refresh(@handle.root["service_action"])
-                @handle.log("info", "Ending Refresh")
+                @handle.log(:info, "Starting Refresh")
+
+                begin
+                  service.refresh(@handle.root["service_action"])
```
mkanoor: @billfitzgerald0120 
Should we test the service_action being passed in?

tinaafitz: @mkanoor @billfitzgerald0120 Madhu and I discussed that all of the generic state machine methods should validate the service_action. We're going to follow up with a PR for the method and spec changes.

PR #50, opened 2017-02-13
Generic Service State Machine - modified update_status Automate method.

tinaafitz:
Changed log message levels from error to info, and added request user message update.


tinaafitz:
@mkanoor Please review.

PR #51, opened 2017-02-13
Generic Service State Machine method update.

tinaafitz:
Validate service action.


tinaafitz:
@mkanoor Please review.

mkanoor:
@tinaafitz 
Can we pass a bogus service_action to this method to test the error scenario

tinaafitz:
@mkanoor - Added bogus service action test.

PR #53, opened 2017-02-14
Automate method to list ansible credentials

mkanoor:
https://www.pivotaltracker.com/story/show/137304515

Automate method to list machine, network and other ansible credentials.
The method accepts a type and yields a
{id => name} hash
e.g.
{id1 => "test1", id2 => "test2"} 

Depends on PR https://github.com/ManageIQ/manageiq/pull/13912

### DIFF
--- a/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
+++ b/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
@@ -34,10 +34,8 @@ module ManageIQ
 
             def fetch_list_data
               credential_list = Hash[*credentials.pluck(:id, :name).flatten]
-              @handle.log(:info, "Number of credentials found #{credential_list.count}")
-              return nil => "<none>" if credential_list.blank?
-              credential_list[nil] = "<select>" if credential_list.length > 1
-              credential_list
+              @handle.log(:debug, "Number of credentials found #{credential_list.count}")
+              {nil => '<Default>'}.merge(credential_list)
             end
 
             def fill_dialog_field(list)
--- a/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
+++ b/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
@@ -8,6 +8,7 @@ module ManageIQ
       module AnsibleTower
         module Operations
           class AvailableCredentials
+            AUTH_CLASS = "ManageIQ_Providers_AutomationManager_Authentication".freeze
             def initialize(handle = $evm)
               @handle = handle
             end
@@ -26,8 +27,9 @@ module ManageIQ
             end
 
             def credentials
-              result = provider.try(:credentials) || []
-              result.select { |c| c.type == @handle.inputs['credential_type'] }
+              @handle.vmdb(AUTH_CLASS).where("resource_id = ? AND type = ?",
+                                             provider.try(:id),
+                                             @handle.inputs['credential_type'])
             end
 
             def fetch_list_data
@@ -42,10 +44,10 @@ module ManageIQ
               dialog_hash = {
                 'sort_by'       => "value",
                 'data_type'     => "string",
-                'required'      => true,
+                'required'      => false,
                 'sort_order'    => "ascending",
                 'values'        => list,
-                'default_value' => list.length == 1 ? list.keys.first : nil
+                'default_value' => nil
               }
 
               dialog_hash.each { |key, value| @handle.object[key] = value }
--- a/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
+++ b/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
@@ -1,0 +1,62 @@
+#
+# Description: provide the dynamic list content from available credentials
+# for Ansible Provider
+#
+module ManageIQ
+  module Automate
+    module AutomationManagement
+      module AnsibleTower
+        module Operations
+          class AvailableCredentials
+            def initialize(handle = $evm)
+              @handle = handle
+            end 
+
+            def main
+              fill_dialog_field(fetch_list_data)
+            end 
+
+            private
+    
+            def get_provider
+              # for provisioning we use     service_template
+              # for reconfig and retire use service
+              service = @handle.root['service_template'] || @handle.root['service']
+              service.try(:job_template, 'Provision').try(:manager)
+            end 
+
+            def get_credentials
+              credentials = get_provider.try(:credentials) || []
+              credentials.select { |c| c.type == @handle.inputs['credential_type'] }
+            end
+            
+            def fetch_list_data
+              credential_list = Hash[*get_credentials.pluck(:id, :name).flatten]
+              @handle.log(:info, "Number of credentials found #{credential_list.keys.count}")
+              return nil => "<none>" if credential_list.blank?
+              credential_list[nil] = "<select>" if credential_list.length > 1
+              credential_list
+            end
+
+            def fill_dialog_field(list)
+              dialog_hash = {
+                'sort_by'       => "value",
+                'data_type'     => "string",
+                'required'      => true,
+                'sort_order'    => "ascending",
+                'values'        => list,
+                'default_value' => list.length == 1 ? list.keys.first : nil
+              }
+
+              dialog_hash.each { |key, value| @handle.object[key] = value }
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  ManageIQ::Automate::AutomationManagement::AnsibleTower::Operations::AvailableCredentials.new.main
+end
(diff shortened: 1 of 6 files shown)


mkanoor:
@bzwei @tinaafitz 
Please review

tinaafitz:
@mkanoor Looks good.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,62 @@
...
+              return nil => "<none>" if credential_list.blank?
+              credential_list[nil] = "<select>" if credential_list.length > 1
+              credential_list
+            end
+
+            def fill_dialog_field(list)
+              dialog_hash = {
+                'sort_by'       => "value",
+                'data_type'     => "string",
+                'required'      => true,
```
bzwei: credential is optional since it is already predefined.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,62 @@
...
+                'sort_order'    => "ascending",
+                'values'        => list,
+                'default_value' => list.length == 1 ? list.keys.first : nil
```
bzwei: The list should be like
```
nil => '<Default>'
id => 'credential name'
```
default will be always `nil => <Default>`, which means to use the one already defined in the job template.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,62 @@
...
+            def provider
+              # for provisioning we use     service_template
+              # for reconfig and retire use service
+              service = @handle.root['service_template'] || @handle.root['service']
+              service.try(:job_template, 'Provision').try(:manager)
+            end
+
+            def credentials
+              result = provider.try(:credentials) || []
+              result.select { |c| c.type == @handle.inputs['credential_type'] }
```
bzwei: you can use `provider.credentials.find_by(:type => type)`. How does `inputs['credential_type']` get set?

mkanoor: @bzwei 
The credential_type is set as an input parameter into the method. So the 2 instances **available_network_credential** will set 
credential_type =
ManageIQ::Providers::AnsibleTower::AutomationManager::NetworkCredential
**available_machine_credentials** will set 
credential_type =
ManageIQ::Providers::AnsibleTower::AutomationManager::MachineCredential

The method itself has a default credential_type set to
ManageIQ::Providers::AnsibleTower::AutomationManager::MachineCredential

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
mkanoor: @bzwei 
I dont understand what the required means in the context of dialogs. I will ask @eclarizio

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
mkanoor: @bzwei 
The find_by works on associations, by the time we get the credentials on the Automate method side it has been converted into an array.

mkanoor: @bzwei 

@handle.vmdb('Authentication').where("resource_id = ? AND type = ?",provider.id, @handle.inputs['credential_type'])

This would work do you prefer this

bzwei: Thanks for the explanation. From performance point of view, this is better.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,64 @@
...
+              fill_dialog_field(fetch_list_data)
+            end
+
+            private
+
+            def provider
+              # for provisioning we use     service_template
+              # for reconfig and retire use service
+              service = @handle.root['service_template'] || @handle.root['service']
+              service.try(:job_template, 'Provision').try(:manager)
```
gmcculloug: @mkanoor It is possible that the different actions could run JobTemplates on different providers.  We would need to determine the `service_action` here and not just hardcode `Provision`.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,64 @@
...
+
+            def credentials
+              @handle.vmdb(AUTH_CLASS).where("resource_id = ? AND type = ?",
+                                             provider.try(:id),
+                                             @handle.inputs['credential_type'])
+            end
+
+            def fetch_list_data
+              credential_list = Hash[*credentials.pluck(:id, :name).flatten]
+              @handle.log(:info, "Number of credentials found #{credential_list.keys.count}")
```
gmcculloug: Any reason why `credential_list.keys.count` and not just `credential_list.count`?

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,64 @@
...
+            end
+
+            def fill_dialog_field(list)
+              dialog_hash = {
+                'sort_by'       => "value",
+                'data_type'     => "string",
+                'required'      => false,
+                'sort_order'    => "ascending",
+                'values'        => list,
+                'default_value' => nil
```
gmcculloug: `sort_by` should be "description" and since the value is an ID `data_type` should be "integer".

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.yaml
```diff
@@ -0,0 +1,32 @@
...
+    location: inline
+  inputs:
+  - field:
+      aetype: 
+      name: credential_type
+      display_name: 
+      datatype: 
+      priority: 1
+      owner: 
+      default_value: ManageIQ::Providers::AnsibleTower::AutomationManager::MachineCredential
```
gmcculloug: Can `default_value` be blank since the instance is always passing this parameter?

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
mkanoor: @gmcculloug I discussed this with @bzwei and this is where we came up with the the restriction that all job templates in a service have to point to the same manager. Since credentials are connected to a manager for the entire job template we will have a single manager. So all service_actions (Provision, Reconfigure, Retire) will point to the same manager and hence the same set of credentials.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
mkanoor: @gmcculloug didn't realize Hash has a count method

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.yaml
mkanoor: @gmcculloug 
One reason I decided to do that was to preserve the expected format of the value when the method gets copied around to a different class/domain without instances.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
bzwei: Good for debugging, but I am not convinced we want to log the count here in final release.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -0,0 +1,64 @@
...
+            def credentials
+              @handle.vmdb(AUTH_CLASS).where("resource_id = ? AND type = ?",
+                                             provider.try(:id),
+                                             @handle.inputs['credential_type'])
+            end
+
+            def fetch_list_data
+              credential_list = Hash[*credentials.pluck(:id, :name).flatten]
+              @handle.log(:info, "Number of credentials found #{credential_list.keys.count}")
+              return nil => "<none>" if credential_list.blank?
```
bzwei: I should simply do 
```
def fetch_list_data
  credential_list = Hash[*credentials.pluck(:id, :name).flatten]
  {nil => '<Default>'}.merge(credential_list)
end
```

mkanoor:
@bzwei @gmcculloug 
Please review

PR #67, opened 2017-03-03
[WIP] Refresh the dest host on VmDeployedEvent

agrare:
We have seen where cloning a VM will emit a `VmDeployedEvent` but not a `CloneVM_Task_Complete` event which leads to a targeted refresh of the dest_host not being queued and the provision request timing out.

https://bugzilla.redhat.com/show_bug.cgi?id=1429032

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/VC.class/vmdeployedevent.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/VC.class/vmdeployedevent.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: VmDeployedEvent
+    inherits: 
+    description: 
+  fields:
+  - rel4:
+      value: "/System/event_handlers/event_action_refresh?target=dest_host"


agrare:
I'm not sure the point of some of the other fields in `CloneVM_Task_Complete` (https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/EVM.class/clonevm_task_complete.yaml#L10-L18) but I assume I should copy those over here?
Also what is the difference between the EVM.class and the VC.class?  `CloneVM_Task_Complete` seems like a vmware specific event.

gmcculloug:
@lfu Please review

agrare:
So to be more accurate, we never got a `CloneVM_Task` event which is why we never created the `CloneVM_Task_Complete` event.

agrare:
>Also what is the difference between the EVM.class and the VC.class?

I see now that all the `EVM.class` events are the ones that we generate so ignore that question :smile:

lfu:
@agrare If VmDeployedEvent is raised from a VM clone, it should call: 
```
/System/event_handlers/event_action_policy?target=src_vm&policy_event=vm_clone&param=host
```
So the policies based on VM clone event will be checked.

Also please follow the steps in [Creating a Pull Request for enhancements to ManageIQ Automate Domain](http://talk.manageiq.org/t/creating-a-pull-request-for-enhancements-to-manageiq-automate-domain/1041) to ensure the format of the yaml file.

agrare:
This appears to have been an issue with the customer VC, closing.

PR #68, opened 2017-03-08
Add Automate modeling for External Ansible Tower Events.

lfu:
https://www.pivotaltracker.com/story/show/140758707

cc @gmcculloug @blomquisg

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/AnsibleTower.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/AnsibleTower.class/__class__.yaml
@@ -1,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: 
+    name: AnsibleTower
+    type: 
+    inherits: 
+    visibility: 
+    owner: 
+  schema:
+  - field:
+      aetype: assertion
+      name: guard
+      display_name: 
+      datatype: 
+      priority: 1
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: attribute
+      name: logical_event
+      display_name: 
+      datatype: 
+      priority: 2
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: method
+      name: on_entry
+      display_name: 
+      datatype: 
+      priority: 3
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: relationship
+      name: rel1
+      display_name: 
+      datatype: 
+      priority: 4
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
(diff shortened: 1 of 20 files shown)


lfu:
@miq-bot assign @gmcculloug
@miq-bot add_label providers/ansible_tower,automate,euwe/no

content/automate/ManageIQ/System/Event/EmsEvent/ANSIBLE_TOWER.class/__class__.yaml
```diff
@@ -0,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: Ansible Tower
+    name: ANSIBLE_TOWER
```
gmcculloug: Suggest
```yaml
display_name: 
name: Ansible_Tower
```

Based on discussion from https://github.com/ManageIQ/manageiq-content/pull/69#pullrequestreview-25906569

gmcculloug:
@bzwei Please review.  Wondering if it makes sense to issue refresh for all of the listed resources.  I think there might be some object types that we do not collect, or we are not concerned with for embedded provider, that may not need to refresh.  For example, groups or hosts.

content/automate/ManageIQ/System/Event/EmsEvent/Ansible_Tower.class/_missing.yaml
```diff
@@ -0,0 +1,10 @@
+---
```
bzwei: @lfu is there a reason to choose `Ansible_Tower.class` rather than `AnsibleTower.class`?

bzwei:
Even for external provider we don't need to refresh on receiving any job create/update/delete events.

content/automate/ManageIQ/System/Event/EmsEvent/Ansible_Tower.class/_missing.yaml
lfu: It has to match what is defined [here](https://github.com/ManageIQ/manageiq/blob/7d7fe087cb0eea8d20ad737ff07f0a4bd7021be1/app/models/manageiq/providers/ansible_tower/automation_manager/event_parser.rb#L5).

gmcculloug: Wondering if the source reference should be changed to be a mixed-case value here and return `AnsibleTower` which would be inline with what is expected.

The original source value was `VC` and the events added after all followed the same uppercase pattern.

Examples:
[app/models/manageiq/providers/openstack/cloud_manager/event_parser.rb#L10](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/openstack/cloud_manager/event_parser.rb#L10)
[app/models/manageiq/providers/google/cloud_manager/event_parser.rb#L14](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/google/cloud_manager/event_parser.rb#L14)
[app/models/manageiq/providers/storage_manager/cinder_manager/event_parser.rb#L12](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/storage_manager/cinder_manager/event_parser.rb#L12)

@lfu Do you know if this would adversely effect anything?

Any thoughts?  @Fryguy @blomquisg

gmcculloug: @lfu Talking to @Fryguy about this and it would be better to follow existing pattern which would mean that:

https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/event_parser.rb#L5
Should be changed to `EMBEDDEDANSIBLE`

https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/ansible_tower/automation_manager/event_parser.rb#L5
Should be changed to `ANSIBLETOWER`

Then we would need to update this PR and https://github.com/ManageIQ/manageiq-content/pull/69 to match the new source changes.  (And use mixed-case in the automate class so it looks nicer like we do for `OpenStack`)

lfu: Sounds good 👍

PR #69, opened 2017-03-08
Add  Automate modeling for Embedded Ansible Events.

lfu:
https://www.pivotaltracker.com/story/show/140758691

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/EmbeddedAnsible.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/EmbeddedAnsible.class/__class__.yaml
@@ -1,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: 
+    name: EmbeddedAnsible
+    type: 
+    inherits: 
+    visibility: 
+    owner: 
+  schema:
+  - field:
+      aetype: assertion
+      name: guard
+      display_name: 
+      datatype: 
+      priority: 1
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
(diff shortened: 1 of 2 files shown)


lfu:
@miq-bot assign @gmcculloug
@miq-bot add_label providers/ansible_tower,automate,euwe/no

lfu:
@miq-bot add_label providers/ansible_tower
@miq-bot add_label automate

content/automate/ManageIQ/System/Event/EmsEvent/EMBEDDED_ANSIBLE.class/__class__.yaml
```diff
@@ -0,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: Ansible Tower
+    name: ANSIBLE_TOWER
```
gmcculloug: Should be:
```yaml
  display_name: Ansible
  name: Embedded Ansible
```

lfu: 👍 Right.
Should we use ```Embedded Ansible``` to tell from external Ansible?
```
    display_name: Embedded Ansible
    name: EMBEDDED_ANSIBLE
```

gmcculloug: I do not think we need to duplicate the information.  Can we use this:
```yaml
  display_name:
  name: Embedded_Ansible
```

The duplications and upper-case make it look odd
image

I would prefer to just have:
image

As for external Ansible that is distinguished by the "Ansible_Tower" name.

bzwei:
I agree we don't need to do refresh on any embedded tower events.
If we must enable the refresh, two scenarios need to be considered:
1. Refresh does not include job at all.
2. ManageIQ playbook service creates/deletes inventory before/after execution. Running refreshes to collect a temporary inventory and delete it later makes no sense.

PR #70, opened 2017-03-09
hostname has been changed to use - instead of _

mkanoor:
Travis build failure, changed spec to use hostname instead of name

Related to this PR https://github.com/ManageIQ/manageiq/pull/13946

### DIFF
--- a/spec/automation/unit/method_validation/scvmm_microsoft_best_fit_least_utilized_spec.rb
+++ b/spec/automation/unit/method_validation/scvmm_microsoft_best_fit_least_utilized_spec.rb
@@ -62,7 +62,7 @@ describe "SCVMM microsoft_best_fit_least_utilized" do
           ws.root
           miq_provision.reload
 
-          expect(miq_provision.options[:placement_host_name]).to eq([host.id, host.name])
+          expect(miq_provision.options[:placement_host_name]).to eq([host.id, host.hostname])
           expect(miq_provision.options[:placement_ds_name]).to   eq([storage.id, storage.name])
         end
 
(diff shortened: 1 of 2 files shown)


spec/automation/unit/method_validation/scvmm_microsoft_best_fit_least_utilized_spec.rb
```diff
@@ -62,7 +62,7 @@
           ws.root
           miq_provision.reload
 
-          expect(miq_provision.options[:placement_host_name]).to eq([host.id, host.name])
+          expect(miq_provision.options[:placement_host_name]).to eq([host.id, host.hostname])
```
gmcculloug: This change is not fixing the issue, just masking it.  

The value set in `:placement_host_name` should be the host `id` and `name`.  But there was a code change to [app/models/mixins/miq_provision_mixin.rb#L293](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/miq_provision_mixin.rb#L293) to support floating_ips which looked at `address` first.

More recently the FloatingIP model was updated to have a name property [app/models/floating_ip.rb#L20](https://github.com/ManageIQ/manageiq/blob/master/app/models/floating_ip.rb#L20) but this should be changed to use `alias_attribute`.

Also, the automate service_model does not seems to honor the `alias_attribute` in my testing which means we would need to add `expose :name` to the FloatingIP service model.

Once the FloatingIP model supports the name the logic in `miq_provision_mixin` should be reduce to always return the name.

Note: For the host object the `address` attributes is aliased to `hostname` and is being depreciated.  https://github.com/ManageIQ/manageiq/blob/master/app/models/host.rb#L96

In summary, this PR should be closed and the follow changes made to the main ManageIQ repo.
1) Expose name on the FloatingIP service model
2) Change name method to an `alias_attribute` on the FloatingIP AR model
3) Modify the `resource_display_name` in `miq_provision_mixin.rb` to remove the `address` check.

gmcculloug:
Closing in favor of https://github.com/ManageIQ/manageiq/pull/14258

PR #73, opened 2017-03-14
Associate VM created from ansible playbook to original service

jeromemarc:
When provisioning a new VM or instance using an Ansible playbook as part of a CloudForms service, the new VM is not associated to the original service. This method provides a way to lookup the original service based on the ansible job id and associates the new VM to the original service. This is called as the last step in the Ansible playbook using the /api/automation_requests CloudForms REST API call.

### DIFF
--- a/content/automate/ManageIQ/System/Request.class/__methods__/add_vm_to_service.rb
+++ b/content/automate/ManageIQ/System/Request.class/__methods__/add_vm_to_service.rb
@@ -1,52 +1,73 @@
 #
-# Description: add_VM_to_Service
+# Description: Add a VM to a Service
+# Input: job_id        The ID of the Job
+#        vm_name       The VM Name
 #
+#
+module ManageIQ
+  module Automate
+    module System
+      module Request
+        class AddVmToService
 
-begin 
-  $evm.log("info", "Called add_VM_to_Service")
-  job_id = $evm.object['job_id']
-  vm_name = $evm.object['vm_name']
-  $evm.log("info", "Called add_VM_to_Service with job_id: #{job_id} and vm_name: #{vm_name}")
-  
-  # Lookup for Service associated with ansible tower job_id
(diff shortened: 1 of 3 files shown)


mkanoor:
@jeromemarc 
For the newer methods we are switching to a new format which helps us to test the methods easily.
For example you can look at https://github.com/ManageIQ/manageiq-content/blob/master/spec/content/automate/ManageIQ/Cloud/Orchestration/Operations/Methods.class/__methods__/available_availability_zones_spec.rb
I have emailed you your updated method which follows that convention. If you can test it in your environment and add back to the PR others would be able to comment on it.

mkanoor:
@jeromemarc 
We are planning on passing extra variables related to ManageIQ to very playbook. It would be in a separate dictionary/hash called ManageIQ.

The variables are
ManangeIQ.MIQ_URL = <<URL Address of our RESTAPI endpoint>
ManageIQ.MIQ_TOKEN = Token that can be used to Authenticate
ManageIQ.MIQ_SERVICE = the service id partial e.g. services/55
ManageIQ.MIQ_REQUESTOR = the user that ordered the service
ManageIQ.MIQ_SERVICE_ACTION = Provisioning|Reconfigure|Retirement

If the service is available to you in the playbook then you don't have to pass in the job_id, the service id would suffice

Also when you call an AutomateRequest you can pass in objects as Service::service=<<id>>
This will fetch an object from the database and allow you to access it as $evm.root['service'] which will further simplify your method.

There is also a way to pass in the VM's as an array so that you can do all of them in a single method call.
Array::vms=10,20,34,23

mkanoor:
@jeromemarc 
If we use the vm array and service the automate method would become much simpler

``` 
vms = $evm.root['vms']
service = $evm.root['service']
vms.each { |vm| vm.add_to_service(service }
```

jeromemarc:
Closed as alternative solution was implemented (using REST API instead of Automate).

PR #79, opened 2017-03-28
Small updates

bdunne:
Minor changes based on `rails plugin new --mountable`

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,14 +1,15 @@
 require 'bundler/setup'
-require 'bundler/gem_tasks'
-require 'manageiq-content'
 
 begin
   require 'rspec/core/rake_task'
 
   APP_RAKEFILE = File.expand_path("../spec/manageiq/Rakefile", __FILE__)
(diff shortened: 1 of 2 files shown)


bdunne:
@Fryguy Please review

Rakefile
```diff
@@ -1,15 +1,16 @@
 require 'bundler/setup'
-require 'bundler/gem_tasks'
```
Fryguy: Why is this removed?

Fryguy:
LGTM :+1:

cc @durandom ... we'll probably want to this in the generator as well as the other plugin repos for consistency

durandom:
> cc @durandom ... we'll probably want to this in the generator as well as the other plugin repos for consistency (and for keeping it as close to a clean rails plugin as possible)

except for the statistics.rake tasks this is already in the generator - but I'm collecting this stuff in my branch https://github.com/ManageIQ/manageiq/compare/master...durandom:provider_generator

PR #82, opened 2017-03-30
[WIP] When a PR to a release branch runs in travis, it shouldn't use master

bdunne:


bdunne:
The release script should update these references

PR #90, opened 2017-04-10
Refactoring cloud orchestration retirement check_removed_from_provider method

pkomanek:
Purpose or Intent
---------------------

Refactoring Cloud/Orchestration/Retirement/StateMachines/check_removed_from_provider.rb method and adding a new spec. This PR is based on the issue bellow.

Links
-------------------
#8

### DIFF
--- a/content/automate/ManageIQ/Cloud/Orchestration/Retirement/StateMachines/Methods.class/__methods__/check_removed_from_provider.rb
+++ b/content/automate/ManageIQ/Cloud/Orchestration/Retirement/StateMachines/Methods.class/__methods__/check_removed_from_provider.rb
@@ -1,21 +1,47 @@
 #
 # Description: This method checks to see if the stack has been removed from the provider
 #
-$evm.root['ae_result'] = 'ok'
-stack = $evm.root['orchestration_stack']
-$evm.log("info", "Checking stack #{stack.try(:name)} removed from provider")
+module ManageIQ
+  module Automate
+    module Cloud
+      module Orchestration
+        module Retirement
+          module StateMachines
+            module Methods
+              class CheckRemovedFromProvider
 
-if stack && $evm.get_state_var('stack_exists_in_provider')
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Cloud/Orchestration/Retirement/StateMachines/Methods.class/__methods__/check_removed_from_provider.rb
```diff
@@ -1,21 +1,48 @@
...
-      $evm.root['ae_result'] = 'retry'
-      $evm.root['ae_retry_interval'] = '1.minute'
+                def initialize(handle = $evm)
+                  @handle = handle
+                end
+
+                def main
+                  @handle.root['ae_result'] = 'ok'
+                  stack = @handle.root['orchestration_stack']
+                  @handle.log("info", "Checking stack #{stack.try(:name)} removed from provider")
```
mkanoor: @pkomanek 
Should this read
```
@handle.log("info", "Checking if stack #{stack.try(:name) has been removed from provider")
```

mkanoor: @pkomanek 
Should we log what the result of the checking was, since retirement takes some time, is it a concern that we will add log messages that indicate that we are in retry mode.
Should we not have any log messages, because other than the name, that log message doesn't reveal anything extra.

I would vote for not having any log messages, in this method since its not conveying much other than adding extra log lines.

simaishi:
@pkomanek @mkanoor Is there a BZ associated with this? Need a BZ to backport to Fine.

gmcculloug:
@simaishi This is a refactoring only so it is unlikely that there would be a BZ.  This is not needed on the Fine branch, it is more of a "nice to have".

PR #93, opened 2017-04-11
Remove FailedSync event from Kubernetes

simon3z:
Removing FailedSync as too verbose and not informative enough on large clusters.
cc @agrare 

Related to https://github.com/ManageIQ/manageiq/pull/14633


gmcculloug:
@simon3z Will containers be moving to a blacklist event filter like other providers?  If so could this be useful in the future?

Also, it is raising a policy event which is defined here:
[db/fixtures/miq_event_definitions.csv#L196](https://github.com/ManageIQ/manageiq/blob/42e8fc031aacc9c09bed83b923d0f973276d91b1/db/fixtures/miq_event_definitions.csv#L196).

If you are removing all processing for this event you need to consider this policy event as well.  Note, it will require a data migration to remove the seeded data from miq_event_definitions.

agrare:
@gmcculloug I'll try to answer that :)
Openshift doesn't support the standard blacklist right now but it needs to move to using it soon.
@simon3z seemed to imply that the POD_FAILEDSYNC event shouldn't be handled in any case since it is more of an internal openshift event.

gmcculloug:
Got it.  Sounds like we should be removing the policy definition as well.  Please open a PR on the main repo to remove the item I referenced above which would include a data migration to remove the seeded data.

PR #113, opened 2017-05-10
Order Ansible Playbook from a Custom Button using a Method

mkanoor:
https://bugzilla.redhat.com/show_bug.cgi?id=1449361

Allows for using a single dialog that is shared between the Custom Button and the Service Template Ordering.

The button needs 2 parameters along with the tie up to the Service Dialog
screen shot 2017-05-09 at 3 14 07 pm

- service_template_name
- hosts                              localhost|vmdb_object|ip1,ip2,ip3

vmdb_object refers to the context of the current object being selected from the button it could be a host or vm. The method would figure out the object to use at runtime.
screen shot 2017-05-11 at 12 03 44 pm

### DIFF
--- a/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb
+++ b/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb
@@ -41,7 +41,6 @@ module ManageIQ
           def vmdb_object_ip
             vmdb_object.try(:ipaddresses).try(:first).tap do |ip|
               if ip.nil?
-                @handle.log(:error, "IP address not specified for vmdb_object")
                 raise "IP address not specified for vmdb_object"
               end
             end
@@ -49,14 +48,12 @@ module ManageIQ
 
           def vmdb_object
             if @handle.root['vmdb_object_type'].nil?
-              @handle.log(:error, "vmdb_object_type missing in root object")
               raise "vmdb_object_type missing in root object"
             end
 
             vmdb_object_type = @handle.root['vmdb_object_type']
 
             if @handle.root[vmdb_object_type].nil?
-              @handle.log(:error, "vmdb_object #{vmdb_object_type} missing in root object")
               raise "vmdb_object #{vmdb_object_type} missing in root object"
             end
             @handle.root[vmdb_object_type]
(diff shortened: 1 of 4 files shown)


mkanoor:
@gmcculloug @tinaafitz 
Please review

gmcculloug:
@mkanoor I can see this quickly expanding to support other resources like hosts and containers.  Instead of overloading the `hosts` property would it make more sense to use a different key name.

We also have the attribute `vmdb_object_type: vm` or `vmdb_object_type: host` that could be used to dynamically determine what object the custom button is running for.

Would it make sense to use the `vmdb_object_type` as the default value if `hosts` is not provided?  Then the `hosts` property would no longer be required from the custom button setup or from the dialog.

mkanoor:
@gmcculloug 
We won't know what hosts parameter is needed for the playbook, is it localhost, vm or something else. The vm basically says that use the passed in vm. So if we key our logic off of vmdb_object_type we will set the host to be always the vm which might not work for all playbooks.

gmcculloug:
I was thinking if someone passed a `hosts` value from the key/value pair of the dialog we would use that.  Otherwise we use the object specified by `vmdb_object_type`.

My point was this method could be used for a custom_button on a vm, host, container or even multi-object in the near future.  And the vmdb_object_type already tells us the target object type to go after in the root object.  I think this method could be re-usable for different objects types and require less configuration.

content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb
```diff
@@ -0,0 +1,89 @@
...
+            else
+              @handle.root['hosts'] || @handle.root['dialog_hosts']
+            end
+          end
+
+          def vmdb_object_ip
+            vmdb_object.try(:ipaddresses).try(:first).tap do |ip|
+              if ip.nil?
+                @handle.log(:error, "IP address not specified for vmdb_object")
+                raise "IP address not specified for vmdb_object"
```
gmcculloug: @mkanoor Is there a reason to log and raise on every error?  Since the message is the same is raise not informative enough?  If both are needed I would suggest creating a method so we do not have to dup every error string.

mkanoor: @gmcculloug 
Yes I think the raise would be good enough since it would show in the Automate log

simaishi:
Fine backport details:

```
$ git log -1
commit bace533c620591b3206fe11f8a6be5bdbd7df080
Date:   Fri May 12 13:10:10 2017 -0400

    Merge pull request #113 from mkanoor/bz_1449361
    
    Order Ansible Playbook from a Custom Button using a Method
    (cherry picked from commit 3220db060e4f35e7908dbc66a1352a38dd0ffb4c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1450485
```

PR #119, opened 2017-05-26
Targeted refresh enhacements for VM import\rename\migration events

pkliczewski:
We need to make sure that we target vm when refreshing after import,
rename or migration events.

Bug:
https://bugzilla.redhat.com/1404357

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/importexport_import_vm.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/importexport_import_vm.yaml
@@ -9,6 +9,6 @@ object:
     description: 
   fields:
   - rel4:
-      value: "/System/event_handlers/event_action_refresh?target=ems"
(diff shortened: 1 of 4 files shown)


pkliczewski:
@miq-bot assign @gmcculloug

pkliczewski:
@borod108 @masayag please review

gmcculloug:
cc @lfu

borod108:
@pkliczewski When migrating the Vm we should refresh the cluster as well, don't we?
So this will work when the changes that add refreshing the cluster as part of the vm targeted refresh are merged?

pkliczewski:
@borod108 We refresh cluster [1] when we target a vm. The change you mentioned use vm as a target rather than ems. 

[1] https://github.com/ManageIQ/manageiq-providers-ovirt/blob/master/app/models/manageiq/providers/redhat/infra_manager/refresh/strategies/api3.rb#L24

pkliczewski:
Migration verified on small env using one vm.

PR #41, opened 2017-02-08
Automate - Service/Generic finish test.

billfitzgerald0120:
Test for new Service Generic StateMachine method.

PR #127, opened 2017-06-08
Add Available_Projects method for dynamic dropdown in container template service dialog

lfu:
The method would collect the available projects for the dynamic dropdown list of projects during service ordering.

Depends on [manageiq #15356](https://github.com/ManageIQ/manageiq/pull/15356).
Depends on [engine #41](https://github.com/ManageIQ/manageiq-automation_engine/pull/41).
Related PR https://github.com/ManageIQ/manageiq/pull/15340

https://www.pivotaltracker.com/story/show/146528931
@miq-bot assign @gmcculloug 
@miq-bot add_label enhancement, services

### DIFF
--- a/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
+++ b/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
@@ -1,0 +1,54 @@
+#
+# Description: provide the dynamic list content from available flavors
+#
+module ManageIQ
+  module Automate
+    module Container
+      module Openshift
+        module Operations
+          class AvailableProjects
+            def initialize(handle = $evm)
+              @handle = handle
+            end
+
+            def main
+              fill_dialog_field(fetch_list_data)
+            end
+
+            private
+
+            def fetch_list_data
+              service = @handle.root.attributes["service_template"] || @handle.root.attributes["service"]
+              projects = service.try(:container_manager).try(:container_projects)
+
+              project_list = {}
+              projects.each { |p| project_list[p.name] = p.name } if projects
+
+              return nil => "<none>" if project_list.blank?
+
+              project_list[nil] = "<select>" if project_list.length > 1
+              project_list
+            end
+
+            def fill_dialog_field(list)
+              dialog_hash = {
+                'sort_by'       => "description",
+                'data_type'     => "string",
+                'required'      => false,
+                'sort_order'    => "ascending",
+                'values'        => list,
+                'default_value' => list.length == 1 ? list.keys.first : nil
+              }
+
+              dialog_hash.each { |key, value| @handle.object[key] = value }
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  ManageIQ::Automate::Container::Openshift::Operations::AvailableProjects.new.main
+end
--- a/spec/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects_spec.rb
+++ b/spec/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects_spec.rb
@@ -1,0 +1,88 @@
+require_domain_file
+
+describe ManageIQ::Automate::Container::Openshift::Operations::AvailableProjects do
+  let(:root_hash) do
+    { 'service_template' => MiqAeMethodService::MiqAeServiceServiceTemplate.find(service_template.id) }
+  end
+  let(:root_object) { Spec::Support::MiqAeMockObject.new(root_hash) }
+  let(:ae_service) do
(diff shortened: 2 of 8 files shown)


content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
```diff
@@ -0,0 +1,54 @@
+#
+# Description: provide the dynamic list content from available flavors
+#
+module ManageIQ
+  module Automate
+    module Container
+      module Openshift
+        module Operations
+          class AvailablePorjects
+            def initialize(handle = $evm)
```
tinaafitz: @lfu Can we add a spec test for this method?
There are examples here:
https://github.com/ManageIQ/manageiq-content/tree/master/spec/content/automate/ManageIQ/Infrastructure/VM/Transform/Import.class/__methods__

lfu:
@tinaafitz I will work on the specs once the backend models are created.

spec/content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    it_behaves_like "#having only default value"
+  end
+
+  context "workspace has container template service template" do
+    let(:service_template) do
+      FactoryGirl.create(:service_template_container_template)
+    end
+    before do
+      allow_any_instance_of(ServiceTemplateContainerTemplate).to receive(:container_template).and_return(container_template)
```
gmcculloug: @lfu Let's see if we can avoid using the `allow_any_instance_of` call.  I believe you can move this logic up into the `let(:service_template)` block and use the object returned by `FactoryGirl` like this:

```ruby
let(:service_template) do
  FactoryGirl.create(:service_template_container_template).tap do |svc_template|
    allow(svc_template).to receive(:container_template).and_return(container_template)
  end
end
```

lfu: @gmcculloug The service template object created here is a different one from what is retrieved in automate. We have to use  ```allow_any_instance_of``` in this case.

gmcculloug: Can you do something like we did here to make ensure the service model is using the same object?  https://github.com/ManageIQ/manageiq-automation_engine/blob/master/spec/engine/miq_ae_method_service/miq_ae_service_methods_spec.rb#L185

lfu:
cc @bzwei

content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
```diff
@@ -0,0 +1,54 @@
...
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  ManageIQ::Automate::Cloud::Orchestration::Operations::AvailableProjects.new.main
```
bzwei: `Container::Openshift`

content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
```diff
@@ -0,0 +1,54 @@
...
+            end
+
+            def main
+              fill_dialog_field(fetch_list_data)
+            end
+
+            private
+
+            def fetch_list_data
+              service = @handle.root.attributes["service_template"]
```
bzwei: should add  `attributes["service"]`

content/automate/ManageIQ/Container/Openshift/Operations/Methods.class/__methods__/available_projects.rb
```diff
@@ -0,0 +1,54 @@
...
+            end
+
+            def fill_dialog_field(list)
+              dialog_hash = {
+                'sort_by'       => "description",
+                'data_type'     => "string",
+                'required'      => false,
+                'sort_order'    => "ascending",
+                'values'        => list,
+                'default_value' => nil
```
bzwei: `default_value` should be the only available value if the list has size 1.

bzwei:
Do we allow selection value to be `nil`? Do you have default project selected at the service template? Otherwise you may not have a project to deploy the template.

lfu:
@bzwei User has to provide either an existing project name or a new project name. Service dialog can specify one field as required, but it does not have the option to say one of two fields is required. 
Service would raise an error if no project is specified for the provisioning.

gmcculloug:
@bzwei See dialog screenshot in PR https://github.com/ManageIQ/manageiq/pull/15340

PR #129, opened 2017-06-15
Exclude orphaned VMs from quota count

branic:
When calculating the number of VMs during quota check archived VMs are excluded from the return count. This PR adds excluding orphaned VMs as well.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/used.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/used.rb
@@ -6,7 +6,7 @@ def consumption(source)
   {
     :cpu                 => source.allocated_vcpu,
     :memory              => source.allocated_memory,
-    :vms                 => source.vms.count { |vm| vm.id unless vm.archived || vm.orphaned },
+    :vms                 => source.vms.count { |vm| vm.id if vm.ems_id },
     :storage             => source.allocated_storage,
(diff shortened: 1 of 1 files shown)


content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/used.rb
```diff
@@ -6,7 +6,7 @@ def consumption(source)
   {
     :cpu                 => source.allocated_vcpu,
     :memory              => source.allocated_memory,
-    :vms                 => source.vms.count { |vm| vm.id unless vm.archived },
+    :vms                 => source.vms.count { |vm| vm.id unless vm.archived || vm.orphaned },
```
gmcculloug: @branic @tinaafitz Instead of checking each method would it be valid to just check `unless vm.ems_id.nil?`.  (Or even better `if vm.ems_id`)

That is the common condition between the two methods:
https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L1269-L1279

branic: using `if vm.ems_id` makes me and is a bit cleaner with less indirection. I'll make the change.

gmcculloug:
@billfitzgerald0120 Can you help provide tests for this change?

gmcculloug:
Closing as changes were incorporated into PR https://github.com/ManageIQ/manageiq-content/pull/132.

PR #144, opened 2017-07-18
host update

pkliczewski:
It is not needed to perform full refresh when we update a host so
this PR changes the target.

Bug-Url:


pkliczewski:
@miq-bot  assign @gmcculloug

pkliczewski:
@gmcculloug  Not sure why travis failed but it seems not to be related to this PR

gmcculloug:
@pkliczewski Was RHV updated to support targeted refresh or does it still always trigger a full refresh?  While I'm not against this change I am curious if it will have any real impact today.

pkliczewski:
@gmcculloug I introduced targeted refresh long time ago. This PR would improve performance during host refresh.

PR #145, opened 2017-07-18
Host install

pkliczewski:
All three events are generated during host installation and it is not
needed to perform full refresh so we prefer to target the specific
host.

Bug-Url:


pkliczewski:
@miq-bot assign @gmcculloug

PR #146, opened 2017-07-18
Host removal

pkliczewski:
It is not needed to run full refresh when a host is removed we can
target it instead.

Bug-Url:


pkliczewski:
@miq-bot assgin @gmcculloug

pkliczewski:
@miq-bot assign @gmcculloug

PR #153, opened 2017-07-28
Target template related actions

pkliczewski:
As part of our performance improvements we want to reduce the scope of a
refresh when adding and removing templates.

Depends on https://github.com/ManageIQ/manageiq-providers-ovirt/pull/70

Bug-Url:
https://bugzilla.redhat.com/1451577

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/template_import.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/template_import.yaml
@@ -9,4 +9,4 @@ object:
     description: 
   fields:
   - rel4:
-      value: "/System/event_handlers/event_action_refresh?target=ems"
+      value: "/System/event_handlers/event_action_refresh?target=src_vm"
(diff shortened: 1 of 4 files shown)


pkliczewski:
This patch depends on https://github.com/ManageIQ/manageiq-providers-ovirt/pull/70 and it needs to be merged first.

pkliczewski:
[Bug 1451577 - [RFE] Targeted refresh or Templates events](https://trello.com/c/FCLCOxwW/181-bug-1451577-rfe-targeted-refresh-or-templates-events)

oourfali:
@gmcculloug can you please review this one?
Thanks!

gmcculloug:
Remove WIP label now that dependent PR is merged.

gmcculloug:
@pkliczewski Were you able to test these changes on a live system?  Is the `src_vm` always available in the db at the time the event is processed?  Before that was not always the case for "New VM" events, but I know RHV creates the basic VM instance based on an event.

cc @lfu

pkliczewski:
@gmcculloug When the object is not there we use `event_action_refresh_new_target` to make sure we create it before the refresh. Yes, I have tested it on my env.

lfu:
LGTM 👍

PR #156, opened 2017-07-31
[WIP] Get the correct group when a user belongs to multiple groups.

lfu:
Need to pass the user's group in to automate when the provision starts.
User's current group might have changed before the provision finishes.
So the user's current group from DB may be different from the user's group when the request is sent into automate.

Part of https://github.com/ManageIQ/manageiq/pull/15696.
Depends on https://github.com/ManageIQ/manageiq-automation_engine/pull/61.

https://bugzilla.redhat.com/show_bug.cgi?id=1467364

@miq-bot add_label bug, euwe/yes, fine/yes


lfu:
With changes in https://github.com/ManageIQ/manageiq-automation_engine/pull/61, this PR is not needed any more.

PR #159, opened 2017-08-02
New refresh method callable from the automate

Ladas:
New refresh method callable from the automate, it tries
to call targeted manager refresh if defined, otherwise fallback to
full refresh.

This is needed for invoking https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_builtin_method.rb#L101

### DIFF
--- a/content/automate/ManageIQ/System/event_handlers.class/__methods__/refresh.yaml
+++ b/content/automate/ManageIQ/System/event_handlers.class/__methods__/refresh.yaml
@@ -1,0 +1,32 @@
+---
+object_type: method
+version: 1.0
+object:
+  attributes:
+    name: refresh
+    display_name: 
+    description: 
+    scope: instance
+    language: ruby
+    location: builtin
+  inputs:
+  - field:
+      aetype: 
+      name: target
+      display_name: 
+      datatype: string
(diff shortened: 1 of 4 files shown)


Ladas:
@cc @gmcculloug @blomquisg we need this so we can invoke the new targeted refresh from the automate

the usage is just "/System/event_handlers/refresh" and the targets are scanned inside, or the result can be full refresh, if we disable the targeted refresh via settings.

gmcculloug:
@lfu Please review

lfu:
@Ladas Does this refresh method work only for Amazon? Or eventually it will work for all providers?
Would like to know if this is a replacement for ```event_action_refresh``` and ```event_action_refresh_sync```.

Ladas:
@lfu we will use it for targeted refresh of AWS and OpenStack, for 4.6. In other providers, it would just invoke full refresh, because of a missing Settings value.

At some point, it will replace ```event_action_refresh```, but that will take some time. (we will need targeted graph refresh in all providers)

lfu:
Changes LGTM 👍

PR #176, opened 2017-08-23
make order_ansible_playbook use floating ip address first if available

ramrexx:


gmcculloug:
@ramrexx Can you add a test for this change here (https://github.com/ManageIQ/manageiq-content/blob/master/spec/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook_spec.rb).

Let us know if you need some help with this.

ramrexx:
@gmcculloug - yes i could use some guidance for this as I have never written a test before.

gmcculloug:
@ramrexx This is not a straightforward as I would have hoped since most of the methods here are private.  I was playing around with some updates to the spec and I'll post back here when I have something.

gmcculloug:
@ramrexx I realized this was built on the `master` branch instead of creating a feature branch for these change.  To fix that and to include my test changes I created a new branch and cherry-picked your commit to it.

Closing in favor of https://github.com/ManageIQ/manageiq-content/pull/195.

PR #180, opened 2017-09-01
Do target scan for every event

Ladas:
Do target scan for every event, leading to targeted
refresh for any target we recognize in the event.

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/_missing.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Amazon.class/_missing.yaml
@@ -7,4 +7,6 @@ object:
     name: ".missing"
     inherits: 
     description: 
-  fields: []
+  fields:
+    - rel4:
+        value: "/System/event_handlers/refresh"


Ladas:
Last comment from @blomquisg in https://github.com/ManageIQ/manageiq-content/pull/26
```
Some of the concern here is that this would have every event from Amazon causing a full refresh of the EMS.

However, now we have the capability to do targeted graph-based refreshes in AWS. So, we might want to revisit this idea. Maybe there should be a way to indicate what type of thing should be updated in the event handler? Then, this could be setup to do a targeted refresh based on the event.

Eventually maybe this could be changed into the streaming event -> inventory updates that @agrare is playing with for VMWare.
```

gmcculloug:
@Ladas Based on your last comment I'm not sure what your intentions are with this PR.

> So maybe I should define few manually for now and this solution (refresh based on every event) would be done later

Can you please clarify if the "few manually" defined events are in PR https://github.com/ManageIQ/manageiq-content/pull/183 and this PR is not intended to be merged at this time?

Ladas:
yeah, lets close this in favor of https://github.com/ManageIQ/manageiq-content/pull/183

PR #182, opened 2017-09-06
Do targeted refresh in Stack check provisioned

Ladas:
Do targeted refresh in Stack check provisioned, if targeted
refresh is allowed.

Needs PRs:
- [x] https://github.com/ManageIQ/manageiq-automation_engine/pull/68
- [x] https://github.com/ManageIQ/manageiq/pull/15936

### DIFF
--- a/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
+++ b/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
@@ -47,10 +47,15 @@ module ManageIQ
 
               def refresh_provider(service)
                 provider = service.orchestration_manager
+                stack    = service.orchestration_stack
 
-                @handle.log("info", "Refreshing provider #{provider.name}")
+                @handle.log("info", "Refreshing provider '#{provider.name}' and stack '#{stack.name}'")
                 @handle.set_state_var('provider_last_refresh', provider.last_refresh_date.to_i)
-                provider.refresh
+                refreshed_stack = @handle.vmdb(:orchestration_stack).find_by(:name    => stack.name,
+                                                                             :ems_ref => stack.ems_ref)
+                if refreshed_stack
+                  refreshed_stack.refresh
+                end
               end
 
               def refresh_may_have_completed?(service)
--- a/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
+++ b/content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
@@ -51,11 +51,7 @@ module ManageIQ
 
                 @handle.log("info", "Refreshing provider '#{provider.name}' and stack '#{stack.name}'")
                 @handle.set_state_var('provider_last_refresh', provider.last_refresh_date.to_i)
-                refreshed_stack = @handle.vmdb(:orchestration_stack).find_by(:name    => stack.name,
-                                                                             :ems_ref => stack.ems_ref)
-                if refreshed_stack
-                  refreshed_stack.refresh
-                end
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
```diff
@@ -47,10 +47,15 @@ def prepare_and_call_refresh_provider(service)
...
               def refresh_provider(service)
                 provider = service.orchestration_manager
+                stack    = service.orchestration_stack
 
-                @handle.log("info", "Refreshing provider #{provider.name}")
+                @handle.log("info", "Refreshing provider '#{provider.name}' and stack '#{stack.name}'")
                 @handle.set_state_var('provider_last_refresh', provider.last_refresh_date.to_i)
-                provider.refresh
+                refreshed_stack = @handle.vmdb(:orchestration_stack).find_by(:name    => stack.name,
+                                                                             :ems_ref => stack.ems_ref)
```
gmcculloug: @bzwei Can you review these changes?  I do not think we are always going to have a `stack` when this is called so the references to `stack.name` and `stack.ems_ref` will raise errors.

bzwei: @Ladas  Here the `stack` object was inserted by our provisioning service. It does not have all the relationships associated, thus requires a refresh. Chances are the object may be deleted by another full refresh. 

My questions are
1. Will a target refresh work if the stack object was not a result of refresh but inserted by our provisioning logic? I guess the answer is yes.
2. If the stack object happened to be deleted by an earlier full refresh, how can you start a target refresh? I assume you will need a full refresh in this case.
3. A full/target refresh may not be needed at all if event driven refresh is already engaged. We simply need to wait until the stack object appears in vmdb with an expected status.

Ladas: 1. so right now this require OrchestrationStack record to exist. I could prepare it as a class method, since all we need for refresh is the ems_ref

2. right, so better would be not to require the record. We can still do targeted refresh, which will create the stack if it exists in the provider.

3. Right, for AWS, we catch the event and refresh stack, but it might be quicker to do a periodic targeted refresh, since the Stack state change event might be slow.

Ladas: ok, let me change it to using OrchestrationStack class method, that way it should work even if we do not receive any event

content/automate/ManageIQ/Cloud/Orchestration/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
```diff
@@ -47,10 +47,11 @@ def prepare_and_call_refresh_provider(service)
 
               def refresh_provider(service)
                 provider = service.orchestration_manager
+                stack    = service.orchestration_stack
 
-                @handle.log("info", "Refreshing provider #{provider.name}")
+                @handle.log("info", "Refreshing provider '#{provider.name}' and stack '#{stack.name}'")
```
gmcculloug: @bzwei Any chance that `stack` could be `nil` here and `stack.name` or the `stack.ems_ref` below could fail?

bzwei: When the state machine comes to refresh I think the stack must exist.

gmcculloug:
@Ladas Please review related test failures.

gmcculloug:
@Ladas dump

Ladas:
@gmcculloug ok, specs fixed, they were testing the old way of refresh only

gmcculloug:
Test failure is unrelated.  

PR create to fix https://github.com/ManageIQ/manageiq-content/pull/207

Ladas:
@gmcculloug awesome, I was just wondering what is the failure :-)

PR #195, opened 2017-10-06
Update order_ansible_playbook to use floating IP first if available

gmcculloug:
When referencing `vmdb_object` check `floating_ip_addresses` first then use the `ipaddresses` property.

### DIFF
--- a/spec/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook_spec.rb
+++ b/spec/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook_spec.rb
@@ -132,6 +132,31 @@ describe ManageIQ::Automate::System::Request::OrderAnsiblePlaybook do
     end
 
     it_behaves_like "order playbook"
+
+    context 'without a floating_ip_addresses method defined' do
+      it_behaves_like "order playbook" do
+        let(:svc_vm) { double(:ipaddresses => []) }
+      end
+    end
+
+    context 'when floating_ip_addresses is empty' do
+      it_behaves_like "order playbook" do
+        let(:svc_vm) { double(:floating_ip_addresses => []) }
+      end
+    end
+
+    context 'with a floating_ip_address' do
+      it_behaves_like "order playbook" do
+        let(:floating_ip) { '2.2.2.94' }
+        let(:svc_vm) { double(:floating_ip_addresses => [floating_ip]) }
+        let(:extra_vars) do
+          { :credential  => nil,
+            :hosts       => floating_ip,
+            'param_var1' => 'A',
+            'param_var2' => 'B' }
+        end
+      end
+    end
   end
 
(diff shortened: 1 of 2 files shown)


gmcculloug:
@bronaghs Can you confirm that this logic make sense?  If an instance has a floating IP address (`vm. floating_ip_addresses`) we should favor that over any IP address (`vm.ipaddresses`) when defining a host for Ansible to contact.

bronaghs:
@gmcculloug @mkanoor 
There wont always be a floating Ip.

@Ladas 
vm.ipaddresses will encompass floating IPs if configured, right? Making this is a safer choice.

Ladas:
@gmcculloug @mkanoor yes this looks correct, floating_ips (we have there also public ips) are the ones most likely accessible from the outside, so we should try the first.

spec/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook_spec.rb
```diff
@@ -132,6 +132,25 @@
     end
 
     it_behaves_like "order playbook"
+
+    context 'with no floating_ip_addresses' do
```
Ladas: @gmcculloug would be nice to have a spec where :floating_ip_addresses method is not defined (I think it's defined only for Cloud Vms objects)

or is that included in ```it_behaves_like "order playbook"``` ?

gmcculloug: The default VM in this spec is `FactoryGirl.create(:vm_vmware, :name => 'fred')` which would not have the `floating_ip_addresses` method defined.  Therefore all previous tests would be running under that condition: `"with no machine credential", "with machine credential", "with vm host"`

It would be better to have an explicit test for that in case someone changes the `vm` factory to be a cloud VM.  I'll add it.

Ladas: awesome, thank you

gmcculloug:
@Ladas @mkanoor Added additional test for `without a floating_ip_addresses method defined`.

gmcculloug:
@mkanoor Please review

gmcculloug:
@mkanoor Thoughts on this PR?

As I recall the other suggestion was to discuss with the provider team about making a new virtual column that would combine floating IPs and ipaddresses into a single array that we could reference instead of having to always perform this check to get the most likely public IP.  Would be needed on several object types, not just vm_or_template.

cc @bronaghs @agrare

ramrexx:
@gmcculloug - where are we on this? It seems to have gone idle.

gmcculloug:
@ramrexx Now is a good time to ping @mkanoor as this PR has stalled.  Some recent comments lead me to believe this approach might be the best solution for now.

PR #197, opened 2017-10-11
Fail request when vm migration fails

pkliczewski:
When vm migration fails it is not reflected in migration request status.

BugUrl:
https://bugzilla.redhat.com/1478518

### DIFF
--- a/content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb
+++ b/content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb
@@ -28,18 +28,27 @@ module ManageIQ
                 task_status = task['status']
                 result = task.statemachine_task_status
 
+                status = task.source.migration_status
+                result = status if status
+
                 @handle.log('info', "CheckMigration returned <#{result}> for state <#{task.state}> and status <#{task_status}>")
 
                 case result
                 when 'error'
+                  task.source.reset_migration_status
                   @handle.root['ae_result'] = 'error'
                   reason = @handle.root['vm_migrate_task'].message
                   reason = reason[7..-1] if reason[0..6] == 'Error: '
                   @handle.root['ae_reason'] = reason
+                when 'failure'
+                  task.source.reset_migration_status
+                  @handle.root['ae_result'] = 'error'
+                  @handle.root['ae_reason'] = "Migration failed"
                 when 'retry'
                   @handle.root['ae_result'] = 'retry'
-                when 'ok'
+                when 'ok' || 'success'
                   # Bump State
+                  task.source.reset_migration_status
                   @handle.root['ae_result'] = 'ok'
                 end
               end
--- a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.rb
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.rb
@@ -1,0 +1,54 @@
+module ManageIQ
+  module Automate
+    module System
+      module Event
+        module EmsEvent
+          module RHEVM
+            class UpdateMigrationStatus
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                update_migration_status(vm)
+              end
+
+              private
+
+              def vm_id
+                event_stream = @handle.root['event_stream']
+                if event_stream.nil?
+                  @handle.log(:error, 'event_stream not found')
+                  raise 'event_stream not found'
+                end
+                event_stream.vm_or_template_id
+              end
+
+              def vm
+                @handle.vmdb('Vm', vm_id).tap do |vm|
+                  if vm.nil?
+                    @handle.log(:error, 'VM object not found')
+                    raise 'VM object not found'
+                  end
+                end
+              end
+
+              def migration_status
+                @handle.root['event_type'] == 'VM_MIGRATION_FAILED_FROM_TO' ? 'failure' : 'success'
+              end
+
+              def update_migration_status(vm)
+                @handle.log(:info, "Updating VM [#{vm.name}] migration status to [#{migration_status}]")
+                vm.set_migration_status(migration_status)
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if __FILE__ == $PROGRAM_NAME
+  ManageIQ::Automate::System::Event::EmsEvent::RHEVM::UpdateMigrationStatus.new.main
+end
--- a/content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/VMMigrate.class/__methods__/update_migration_status.rb
+++ b/content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/VMMigrate.class/__methods__/update_migration_status.rb
@@ -47,6 +47,8 @@ module ManageIQ
                   if status == 'migration_complete'
                     task.finished('VM Migrated Successfully')
                   end
+                elsif @handle.root['ae_result'] == 'error'
+                  task.finished('VM migration failed')
                 end
               end
             end
--- a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.yaml
@@ -1,0 +1,12 @@
+---
+object_type: method
+version: 1.0
+object:
+  attributes:
+    name: update_migration_status
+    display_name:
+    description:
+    scope: instance
+    language: ruby
+    location: inline
+  inputs: []
--- a/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/vm_migration_failed_from_to.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/vm_migration_failed_from_to.yaml
@@ -3,14 +3,12 @@ object_type: instance
 version: 1.0
 object:
   attributes:
-    display_name: 
-    name: VM_MIGRATION_DONE
-    inherits: 
-    description: 
+    display_name:
+    name: VM_MIGRATION_FAILED_FROM_TO
+    inherits:
+    description:
   fields:
   - meth4:
       value: update_migration_status
   - rel5:
       value: "/System/event_handlers/event_action_refresh?target=src_vm"
-  - rel6:
-      value: "/System/event_handlers/event_action_policy?target=src_vm&policy_event=vm_migrate&param="


content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.rb
```diff
@@ -0,0 +1,34 @@
...
+
+              def main
+                update_migration_status
+              end
+
+              private
+
+              def update_migration_status
+                if @handle.root['event_type'] == 'VM_MIGRATION_FAILED_FROM_TO'
+                  @handle.root['ae_result'] = 'error'
```
pkliczewski: @tinaafitz This line do not mark migration request as failed. Can you help me to understand how to fail it when `VM_MIGRATION_FAILED_FROM_TO` arrives from the provider?

tinaafitz: @pkliczewski The issue here is that the VM Migration state machine and event processing are entirely separate.  

The vm migrate state machine is should fail on the execute:  
https://github.com/tinaafitz/manageiq/blob/master/app/models/vm_migrate_task.rb#L65

Instead of marking the status as migrated:
https://github.com/tinaafitz/manageiq/blob/master/app/models/vm_migrate_task.rb#L72

tinaafitz: @pkliczewski Do you have an environment setup where I can reproduce the migration failure?

pkliczewski: It seems that we need to do it as part of state machine. Do you have any suggestions or examples how to?

bzwei: @pkliczewski you can mimic 
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_vm_import_status.rb#L42 where you can set a customer attribute by assigning the migration result to a unique key. Don't fail the request here regardless the migration result.

Then you can mimic https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Infrastructure/VM/Transform/StateMachines/VmImport.class/__methods__/wait_for_vm_import.rb#L19 where they check the `import_status` to decide to retry, complete, or fail the `wait_for_vim_import` step. You should do the same in https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb#L27

Lastly, as @tinaafitz mentioned earlier, in https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_migrate_task.rb#L72 change the state from `migrated` to `migrating`. 

Let's see how it works. Later we can work together to find out how it can apply to VMware too.

pkliczewski:
@bzwei Please review. The request status is not updated after running above changes.
I updated [here](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_migrate_task.rb#L72) with `:state => 'active'` because only following values are allowed:
   
Validation failed: State should be pending, finished, migrated, active, queued  Method:[block in method_missing]

bzwei:
@pkliczewski together with the change of `:state => 'active'` were you able to have a full end-to-end run that properly fail the request when migration fails; and complete the request when migration succeeds?

pkliczewski:
@bzwei With this approach I was able to see that the failure was properly passed by the requested stayed in `Active` state. I wonder what am I missing here to have it updated.

bzwei:
@pkliczewski I don't see why the status does does not update. We can schedule a meeting so that you can demo to us.

content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb
```diff
@@ -36,9 +39,12 @@ def checkmigration(task)
                   reason = @handle.root['vm_migrate_task'].message
                   reason = reason[7..-1] if reason[0..6] == 'Error: '
                   @handle.root['ae_reason'] = reason
+                when 'failure'
+                  @handle.root['ae_result'] = 'error'
+                  @handle.root['ae_reason'] = "Migration failed"
                 when 'retry'
```
bzwei: you need retry when result is 'active'

pkliczewski: This would mean that we need to retry from [here](https://github.com/ManageIQ/manageiq-content/pull/197/files#diff-04a70eac4f0eb3c42634d50eafda6dfcR49) which seems not the best idea.

pkliczewski: or from the line you commented

bzwei:
@pkliczewski for the build failure, you can follow the instruction to run
```
bundle exec rake clean_yaml_files
```

bzwei:
The task class https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_migrate_task.rb needs to implement something similar to https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_provision.rb#L69. This method ensured the final state is `finished` and status is either `Ok` or `Error`

pkliczewski:
@bzwei Do you suggest to add `after_ae_delivery` to [vm_migrate_task](https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_migrate_task.rb)?

pkliczewski:
This PR depends on https://github.com/ManageIQ/manageiq/pull/16548

pkliczewski:
[Bug 1478518 - CFME reports VM migration passed when it fails on RHV side](https://trello.com/c/Tti4EKip/215-bug-1478518-cfme-reports-vm-migration-passed-when-it-fails-on-rhv-side)

pkliczewski:
This PR depends on https://github.com/ManageIQ/manageiq-automation_engine/pull/126

pkliczewski:
This PR depends on https://github.com/ManageIQ/manageiq/pull/16548

pkliczewski:
@miq-bot add_labels gaprindashvili/yes

pkliczewski:
@miq-bot add_labels fine/yes

content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb
```diff
@@ -28,18 +28,27 @@ def checkmigration(task)
                 task_status = task['status']
                 result = task.statemachine_task_status
 
+                status = task.source.custom_get(:migration_status)
```
bzwei: The idea is to have a new `Vm#migration_status` method for all Vm types, basically RHEV and VMWare right now. For RHEV implementation you may use a customer attribute to track the status. 
You can set migration_status to `migrating` in the beginning, and `failed` or `succeeded` in the end.
But you don't directly get/delete customer attribute here.
You should not need  https://github.com/ManageIQ/manageiq-automation_engine/pull/126
and https://github.com/ManageIQ/manageiq-providers-ovirt/pull/168

For the new method `Vm#migration_status` you may want to return an array, first being the status, the second to be a detailed message explaining how it fails.

pkliczewski: The main goal for having [1] is to manage how state of `VmMigrateTask` is managed. At the moment what we need is different behavior for RHV so having generic behavior for everything and specific for RHV so [1] seems like good idea. 

Do you suggest to have additional column or use `CustomAttirbute` for which status is stored and retrieved by `Vm#migration_status`?

[1] https://github.com/ManageIQ/manageiq-providers-ovirt/pull/168

bzwei: We don't need [1] but have the initial status `active` for all providers, exactly the change you made before.
We will not add any additional column. The state machine retry (`checkprovision`) will keep on pulling `Vm#migration_status`. In its implementation you can read from custom attribute or directly query the event_streams table as we discussed earlier.

pkliczewski: This is how I understood your point about vmware and `vm.migrate` being done after returning from the call. With current changes vmware code would work and rhv would change the initial status to `active`.

Directly quering `event_streams` would mean that we need to depend on timestamps which is very error prone. We can move `custom_attribute` implementation to `vm` class.

bzwei: We want to change VMWare `vm.migrate` to return immediately and we will rely on `checkmigration` to wait for completion.

VMWare `vm.migrate` calls https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager.rb#L229
 and eventually https://github.com/ManageIQ/vmware_web_service/blob/9497671d9405c61fd3ab35603cab1b738a74986f/lib/VMwareWebService/MiqVimVm.rb#L196. Here we want to add `wait = true` in the argument list like other similar methods in this class. Make sure `vm.migrate` calls it with `wait = false`.

Then in `vm.migration_status` you can implement something similar to https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/provision/cloning.rb#L2

We need to think a way to pass around the `task_mor`, initially returned from `vm.migrate` method.

pkliczewski: OK so it is something different than I understood from our conversation. I will change.

pkliczewski: btw, how do you want to merge the code? Current changes (we need to add `vm.migration_status`) do not affect how other providers work. Do you want to wait on vmware migrate async code before merging this? I think it would be better to have this in and do not affect vmware migration and later when ready it could be merged.

bzwei: Adding `Vm#migration_status` in each provider can be merged first because they are new methods. Adding a flag (`wait = true`) in vmware_web_service can be merged early too.

pkliczewski: I can create `migration_status` in `VmOrTemplate` which should be visible to all the providers. I ask about migration state machine changes which will break vmware migration.

bzwei: state machine change should be merged last. We do not need to add `migration_status` immediately in `VmOrTemplate` because it will be an abstract method anyway. It is good to have but not required in runtime. Can be added later.

pkliczewski: why not to fix rhv issue and keep the current functionality for other providers and later work on vmware migration? Why do we need to split it?

bzwei: If you don't want to make changes to how VMWare migration works now, the minimum we can do is to implement VMWare's `migration_status` to always return `succeeded`.

pkliczewski: OK, I see where we misunderstood each other. I will go with `succeeded` for vmware.

content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.rb
```diff
@@ -0,0 +1,54 @@
...
+                end
+              end
+
+              def migration_status
+                @handle.root['event_type'] == 'VM_MIGRATION_FAILED_FROM_TO' ? 'failure' : 'success'
+              end
+
+              def update_migration_status(vm)
+                @handle.log(:info, "Updating VM [#{vm.name}] migration status to [#{migration_status}]")
+                vm.set_migration_status(migration_status)
```
bzwei: `set_migration_status` method can be exposed in RHEV's VM class, but not the base class.

content/automate/ManageIQ/Infrastructure/VM/Migrate/StateMachines/Methods.class/__methods__/checkmigration.rb
```diff
@@ -28,18 +28,27 @@ def checkmigration(task)
...
                 result = task.statemachine_task_status
 
+                status = task.source.migration_status
+                result = status if status
+
                 @handle.log('info', "CheckMigration returned <#{result}> for state <#{task.state}> and status <#{task_status}>")
 
                 case result
                 when 'error'
+                  task.source.reset_migration_status
```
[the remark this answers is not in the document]
pkliczewski: We use status to indicate whether the process failed or succeeded. We need `reset` to make sure that when next time we migrate the same vm the process is not concluded with previous state.

content/automate/ManageIQ/System/Event/EmsEvent/RHEVM.class/__methods__/update_migration_status.rb
pkliczewski: If we want to not to have default implementation we need to check whether the method exists.

pkliczewski:
https://github.com/ManageIQ/manageiq-providers-ovirt/pull/182

PR #201, opened 2017-10-24
Add policy event triggers for Hawkular

tumido:
Create mapping for Hawkular events so they trigger Middleware events via automate. 

Depends on: https://github.com/ManageIQ/manageiq/pull/16134
Required by: https://github.com/ManageIQ/manageiq-providers-hawkular/pull/71
Resolves: https://issues.jboss.org/browse/JMAN4-200

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.error.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.error.yaml
@@ -3,7 +3,7 @@ object_type: instance
 version: 1.0
 object:
   attributes:
-    display_name: Datasource Error
+    display_name: Datasource Deploy Error
     name: hawkular_datasource.error
     inherits: 
     description: 
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.error.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.error.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: Datasource Error
+    name: hawkular_datasource.error
+    inherits: 
+    description: 
+  fields:
+  - rel4:
+      value: "/System/event_handlers/event_action_policy?target=middleware_server&policy_event=middleware_datasource_error&param="
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.ok.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.ok.yaml
@@ -1,0 +1,14 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: Datasource Success
+    name: hawkular_datasource.ok
+    inherits: 
+    description: 
+  fields:
+  - rel4:
+      value: "/System/event_handlers/event_action_refresh?target=ems"
+  - rel5:
+      value: "/System/event_handlers/event_action_policy?target=middleware_server&policy_event=middleware_datasource_ok&param="
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.ok.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.ok.yaml
@@ -3,7 +3,7 @@ object_type: instance
 version: 1.0
 object:
   attributes:
-    display_name: Datasource Success
+    display_name: Datasource Deploy Success
     name: hawkular_datasource.ok
     inherits: 
     description: 
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_deployment_remove.ok.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_deployment_remove.ok.yaml
@@ -1,0 +1,14 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: Deployment Remove Success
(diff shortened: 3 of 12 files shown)


tumido:
cc @israel-hdez, so after all ([comment](https://github.com/ManageIQ/manageiq-providers-hawkular/pull/71#issuecomment-334815536)) the automate is actually needed to trigger the policy events... ;)

cc @lfu is this how it should look like?
cc @abonas, this is gonna solve the issue with events I ran into, we're moving!

lfu:
@tumido Forgot to mention that we need to follow the steps in [Creating a Pull Request for enhancements to ManageIQ Automate Domain](http://talk.manageiq.org/t/creating-a-pull-request-for-enhancements-to-manageiq-automate-domain/1041) to keep the YAML files format consistent.

content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.error.yaml
```diff
@@ -0,0 +1,14 @@
...
+  attributes:
+    display_name: Datasource Error
+    name: hawkular_datasource.error
+    inherits:
+    description:
+  fields:
+  - rel4:
+      value: "/System/event_handlers/event_action_refresh?target=ems"
+  - rel5:
+      value: "/System/event_handlers/event_action_policy?target=middleware_server&policy_event=mw_datasource_error&param="
```
lfu: @tumido ```policy_event``` raised in this PR does not match what is defined in [your PR for MiqEventDefinition](https://github.com/ManageIQ/manageiq/pull/16134/files#diff-da17f6bf8d1c54c7878221e8110a0577R77).

tumido:
@lfu, it should be fixed now ;)

abonas:
@miq-bot add_label providers/middleware

abonas:
1. what are the sections rel4 and rel5 for? 
2. what is the functionality that is enabled by these files? it seems a lot of overhead to create and maintain  those files per every single event. (or I am missing something?)
(the above are mainly questions to core team)
cc @agrare

gmcculloug:
These files are defining the actions taken when an event is sent into automate.  rel4 is sending a refresh to the provider and rel5 is raising the corresponding policy event for the provider event that the file is handling.

The files are part of the export format for the automate model.

abonas:
@gmcculloug does this mean that it *always* sends a request to refresh when this event arrives?
@tumido have you considered if rel4 is necessary in all cases?

gmcculloug:
Yes, `event_action_refresh?target=ems` is requesting a full provider refresh each time the event is processed.

tumido:
@abonas, yes I think the refresh is necessary, but I would like to hear @israel-hdez opinion on that matter. But I generally think, since the event is notifying about some action on a MiddlewareServer, it's worth refreshing to see update inventory if for example the removal of deployment failed etc... But I'm not 100% sure about that. So if you consider the refresh is not necessary in some cases, I don't mind to remove it.

The "rel4", "rel5" are just slots taken by the action. There are multiple slots available and they are executed in the name order. So if you have defined "rel1", "rel2" and "rel4" they are execuded in order 1,2,4, etc. User can modify the mapping in ManageIQ and put custom actions in the empty slots. so using "rel4" and rel5" allows user to put his custom actions before and after our actions.

abonas:
@tumido my concern is that refresh is a heavy action. and might not be required for all the cases. hence it will be good to review each case separately in that context.

tumido:
@abonas, yes I agree. @israel-hdez has much better grip on what's going on under the hood when such event happens, so he know better. I don't feel educated enough to decide. To me it seems doing a refresh is safer and keeps us consistent if we're not sure...

israel-hdez:
@tumido I would skip the refresh for the error events. I think it's not needed, since we know nothing was added/removed from inventory.
Also, the JDBC drivers aren't inventoried; they are fetched directly from Hawkular when required. The refresh for those cases surely is not needed.

tumido:
@israel-hdez done, please review

content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_datasource.ok.yaml
```diff
@@ -0,0 +1,14 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: Datasource Success
```
abonas: what Datasource Success means? what kind of event is this? I think I'm missing a verb here :)

content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_jdbc.ok.yaml
```diff
@@ -0,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: JDBC Success
```
abonas: missing a verb/action here. what does it mean JDBC success?

tumido: I've tried to follow what we already have [here](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_deployment.ok.yaml) (I considered Deployment as a noun not a verb, my bad). I'm gonna change it so we're more consistent with the https://github.com/ManageIQ/manageiq/pull/16134.

tumido:
@abonas, is this better?

content/automate/ManageIQ/System/Event/EmsEvent/Hawkular.class/hawkular_deployment_remove.ok.yaml
```diff
@@ -3,7 +3,7 @@ object_type: instance
 version: 1.0
 object:
   attributes:
-    display_name: Deployment Remove Success
+    display_name: Application Uneploy Success
```
abonas: typo, missing "d" in undeploy

PR #204, opened 2017-10-26
Nuage events callbacks for targeted refresh

miha-plesko:
With this commit we first specify automate class for dealing with Nuage events. Then we define rel4 callbacks for events that we currently support. All callbacks simply point to the event_target_parser where appripriate targets get queued for refresh.

Related PR where `event_target_parser.rb` implementation is: https://github.com/ManageIQ/manageiq-providers-nuage/pull/28

@miq-bot assign @gmcculloug 

/cc @gberginc @Ladas

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/NUAGE.class/_missing.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/NUAGE.class/_missing.yaml
@@ -1,0 +1,10 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: ".missing"
+    inherits: 
+    description: 
+  fields: []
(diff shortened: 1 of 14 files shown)


content/automate/ManageIQ/System/Event/EmsEvent/NUAGE.class/_missing.yaml
```diff
@@ -0,0 +1,12 @@
...
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: ".missing"
+    inherits: 
+    description: 
+  fields:
+  - rel4:
+      value: "/System/event_handlers/refresh"
```
Ladas: so before @gmcculloug and @blomquisg were rather for manually defining handlers for every event, instead of method_missing. Since the handlers also solve as documentation of 'these are the events we support'

miha-plesko:
@Ladas thanks, I've now added the event definitions for events currently supported by targeted refresh, and removed the rel4 from the _missing.yaml.

chargio:
👌

gmcculloug:
@lfu Please review

lfu:
LGTM 👍

PR #206, opened 2017-10-26
Ability to retire a service from service dialog

psachin:
As per[1] we can retire a service but including 'service_retire_on' in
service dialog still needs customization in Automate. This is
discussed in [2].

I think ManageIQ Domain should handle 'service_retire_on' dialog
option by default. This will save creating a custom domain just to
fetch 'service_retire_on' value from service catalog.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1300428
[2] http://talk.manageiq.org/t/services-retirement-date/786/2

Closes: https://bugzilla.redhat.com/show_bug.cgi?id=1506712


gmcculloug:
Closing in favor of PR https://github.com/ManageIQ/manageiq/pull/16799

PR #207, opened 2017-10-30
Use proper request_task subclass for provisioning spec.

gmcculloug:
Only the ServiceTemplateProvisionTask model exposes the provision_priority method.

Required now that https://github.com/ManageIQ/manageiq/pull/16299 references the `provision_priority` property of ServiceTemplateProvisionTask.

Causing test failures.  Ref https://travis-ci.org/ManageIQ/manageiq-content/jobs/294782150


gmcculloug:
@bzwei Please review.

PR #243, opened 2018-01-19
Add event state machine for refresh.

lfu:
This event state machine replaces the builtin method event_action_refresh_sync that waits for the completion of refresh and hence blocks the worker. 

Part of https://github.com/ManageIQ/manageiq/pull/16868.
https://bugzilla.redhat.com/show_bug.cgi?id=1534631

@miq-bot add_label bug,  fine/yes, gaprindashvili/yes

cc @mkanoor @gmcculloug @tinaafitz

### DIFF
--- a/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
+++ b/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
@@ -1,0 +1,64 @@
+#
+# Description: This method checks the refresh completion.
+#
+
+module ManageIQ
+  module Automate
+    module System
+      module Event
+        module StateMachines
+          module Refresh
+            class CheckRefreshed
+              STATE_FINISHED = 'Finished'.freeze
+              STATUS_OK = 'Ok'.freeze
+
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                check_status(refresh_task)
+              end
+
+              private
+
+              def check_status(task)
+                case task.state
+                when STATE_FINISHED
+                  if task.status == STATUS_OK
+                    @handle.root['ae_result'] = 'ok'
+                  else
+                    @handle.root['ae_result'] = 'error'
+                    @handle.log(:error, "Refresh task ended with error: #{task.message}")
+                  end
+                else
+                  @handle.root['ae_result'] = 'retry'
+                  @handle.root['ae_retry_interval'] = '1.minute'
+                end
+              end
+
+              def refresh_task
+                task_id = @handle.get_state_var(:refresh_task_id).first
+                @handle.log(:info, "Stored refresh task ID: [#{task_id}]")
+                fetch_task(task_id)
+              end
+
+              def fetch_task(task_id)
+                @handle.vmdb(:miq_task).find_by(:id => task_id).tap do |task|
+                  if task.nil?
+                    @handle.log(:error, "Refresh task with id: #{task_id} not found")
+                    raise "Refresh task with id: #{task_id} not found"
+                  end
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
+  ManageIQ::Automate::System::Event::StateMachines::Refresh::CheckRefreshed.new.main
+end
--- a/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/target_refresh.rb
+++ b/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/target_refresh.rb
@@ -1,0 +1,51 @@
+#
+# Description: This method starts a synchronous refresh and save the
+#              task id in the state variables so we can use it when
+#              waiting for the refresh to finish.
+#
+
+module ManageIQ
+  module Automate
+    module System
+      module Event
+        module StateMachines
+          module Refresh
+            class TargetRefresh
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                task_id = event.refresh(refresh_target, true)
+                raise "Refresh task not created" if task_id.blank?
+                @handle.set_state_var(:refresh_task_id, task_id)
+              end
+
+              private
+
+              def event
+                @handle.root["event_stream"].tap do |event|
+                  if event.nil?
+                    @handle.log(:error, 'Event object is nil')
+                    raise 'Event object not found'
+                  end
+                end
+              end
+
+              def refresh_target
+                @handle.object["refresh_target"].tap do |target|
+                  @handle.log(:info, "Refresh target: [#{target}]")
+                  raise "Refresh target not found" if target.nil?
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
+  ManageIQ::Automate::System::Event::StateMachines::Refresh::TargetRefresh.new.main
+end
--- a/content/automate/ManageIQ/System/Event/EmsEvent/EVM.class/reconfigvm_task_complete.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/EVM.class/reconfigvm_task_complete.yaml
@@ -9,6 +9,4 @@ object:
     description: 
   fields:
   - rel4:
-      value: "/System/event_handlers/event_action_refresh_sync?target=src_vm"
-  - rel5:
-      value: "/System/event_handlers/event_action_policy?target=src_vm&policy_event=vm_reconfigure&param="
+      value: "/System/Event/StateMachines/Refresh/default?refresh_target=src_vm&policy_event=vm_reconfigure&policy_target=src_vm"
--- a/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__class__.yaml
@@ -1,0 +1,133 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: Refresh State Machine
+    display_name: 
+    name: Refresh
+    type: 
+    inherits: 
+    visibility: 
+    owner: 
+  schema:
+  - field:
+      aetype: state
+      name: pre
+      display_name: 
+      datatype: string
+      priority: 1
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: ''
+      on_exit: ''
+      on_error: ''
+      max_retries: '100'
+      max_time: 
+  - field:
+      aetype: state
+      name: refresh
+      display_name: 
+      datatype: string
+      priority: 2
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: '100'
+      max_time: 
+  - field:
+      aetype: state
+      name: check_refreshed
+      display_name: 
+      datatype: string
+      priority: 3
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: '100'
+      max_time: 
+  - field:
+      aetype: state
+      name: post
+      display_name: 
+      datatype: string
+      priority: 4
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: ''
+      on_exit: ''
+      on_error: ''
+      max_retries: '100'
+      max_time: 
+  - field:
+      aetype: state
+      name: policy
+      display_name: 
+      datatype: string
+      priority: 5
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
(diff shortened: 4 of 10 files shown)


lfu:
screen shot 2018-01-19 at 11 04 13 am

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/target_refresh.rb
```diff
@@ -0,0 +1,50 @@
...
+                    raise 'Event object not found'
+                  end
+                end
+              end
+
+              def refresh_target
+                target = @handle.object["refresh_target"]
+                @handle.log(:info, "Refresh target: [#{target}]")
+                raise "Refresh target not found" if target.nil?
+                target
```
gmcculloug: You don't need to do this but since you used `tap` above this feels like another example.

```ruby
@handle.object["refresh_target"].tap do |target|
  @handle.log(:info, "Refresh target: [#{target}]")
  raise "Refresh target not found" if target.nil?
end
```

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,61 @@
...
+              end
+
+              def fetch_task(task_id)
+                @handle.object.attributes.sort.each { |k, v| @handle.log(:info, "#{k}: #{v}") }
+                task = @handle.vmdb(:miq_task).find(task_id)
+                if task.nil?
+                  @handle.log(:error, "Refresh task with id : #{task_id} not found")
+                  exit(MIQ_ERROR)
+                end
+                task
```
gmcculloug: Same here with `tap`.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,61 @@
...
+                when 'Finished'
+                  @handle.root['ae_result'] = 'ok'
+                else
+                  @handle.root['ae_result'] = 'retry'
+                  @handle.root['ae_retry_interval'] = 1.minute
+                end
+              end
+
+              def refresh_task
+                task_id = @handle.get_state_var(:task_id).try(:first)
```
gmcculloug: Suggest using a more descriptive key name in place of `:task_id`, like `:refresh_task_id`.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,61 @@
...
+
+              def main
+                check_status(refresh_task)
+              end
+
+              private
+
+              def check_status(task)
+                case task.state
+                when 'Finished'
```
gmcculloug: Move this string into a constant: `STATE_FINISHED = 'Finished'.freeze`

mkanoor:
@lfu 
The MiqTask is pretty generic so we should write the builtin method in a way that it can be reused by other methods that create a MiqTask. The task_id can be either passed into the builtin method or it could directly access some well know state var

lfu:
@mkanoor ```MiqTask``` is returned by ```EmsRefresh.queue_refresh```. Event handler in this case does not create the task.

jrafanie:
@lfu can you add this BZ to the PR?  Thanks!

https://bugzilla.redhat.com/show_bug.cgi?id=1534631

It seems to go together with my change for [the authentications](https://github.com/ManageIQ/manageiq/pull/16864).

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,63 @@
...
+                @handle.log(:info, "Stored refresh task ID: [#{task_id}]")
+                if task_id.nil?
+                  @handle.log(:error, 'Refresh task id not found')
+                  raise 'Refresh task id not found'
+                end
+                fetch_task(task_id)
+              end
+
+              def fetch_task(task_id)
+                @handle.object.attributes.sort.each { |k, v| @handle.log(:info, "#{k}: #{v}") }
```
mkanoor: @lfu 
Is this just for logging purpose? Do we need it?

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,63 @@
...
+
+              def main
+                check_status(refresh_task)
+              end
+
+              private
+
+              def check_status(task)
+                case task.state
+                when STATE_FINISHED
```
mkanoor: @lfu What about the error scenario when the task is finished but has an error? will that get flagged as a retry

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
lfu: @mkanoor This is for information purpose. It would be useful if anything goes wrong. I would prefer to print it out for our good.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
lfu: @mkanoor A ```MiqTask's```  state may be one of ```Initialized, Queued, Active and Finished```.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
gmcculloug: We can get this information from the automation URL in the log already (in a less user-friendly format).  This adds more clutter to an already verbose log.  I would advise dropping it.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
gmcculloug: @mkanoor There is a separate `status` column for capturing that.  [app/models/miq_task.rb#L8-L12](https://github.com/ManageIQ/manageiq/blob/d992f980a857db789c8aa8b3636bf1b6559ca539/app/models/miq_task.rb#L8-L12)

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,62 @@
...
+                  @handle.root['ae_result'] = 'ok'
```
mkanoor: @lfu If the State == "Finished" but Status != "Ok" we should set the ae_result as error and fail the state machine. Otherwise if there is a password failure during refresh we will continue processing the next state which is going to run the policy action.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,61 @@
...
+                if task_id.nil?
+                  @handle.log(:error, 'Refresh task id is nil')
+                  raise 'Refresh task id is nil'
+                end
+                fetch_task(task_id)
+              end
+
+              def fetch_task(task_id)
+                @handle.vmdb(:miq_task).find(task_id)
+              rescue MiqAeException::ServiceNotFound
```
lfu: @mkanoor @tinaafitz @gmcculloug Please review this change.
``` @handle.vmdb(:miq_task).find(task_id)``` would throw an exception instead of returning nil when task is not found.

Comment (gmcculloug) [reply content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb]: Use `@handle.vmdb(:miq_task).find_by(:id => task_id)`

File: content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb [hunk 9]
```diff
@@ -0,0 +1,63 @@
...
+                when STATE_FINISHED
+                  @handle.root['ae_result'] = task.status == STATUS_OK ? 'ok' : 'error'
+                else
+                  @handle.root['ae_result'] = 'retry'
+                  @handle.root['ae_retry_interval'] = '1.minute'
+                end
+              end
+
+              def refresh_task
+                task_id = @handle.get_state_var(:refresh_task_id).try(:first)
```
mkanoor: @lfu what is the order of the task_ids, could we check that both the tasks are done instead of just one. This code only seems to be testing for 1 task id.

mkanoor: @lfu when we store the task ids we might want to store it as an array in the state vars and when we check we would have to validate that all the tasks have completed successfully.

gmcculloug: Processing as an array is a good idea but we could ensure it is an array here if the common use case is going to be a single task id.

mkanoor: @gmcculloug @lfu 
The earlier code was doing this
      task_ids.each { |tid| MiqTask.wait_for_taskid(tid) } if sync
It was handling an array in https://github.com/ManageIQ/manageiq/blob/d992f980a857db789c8aa8b3636bf1b6559ca539/app/models/ems_event/automate.rb#L23

gmcculloug: I see now, you are dealing with multiple miq_task objects coming back from the backend provider call.  👍

lfu: @mkanoor @gmcculloug It is true that the ```task_ids``` returned back from refresh is in an array. [Refresh](https://github.com/ManageIQ/manageiq/blob/962af5d7a934d8d286ac2f938a5d5e05c31efd8f/app/models/ems_refresh.rb#L56) would create multiple ```MiqTasks``` based on the number of ```ems``` these refresh targets belong to.  
In our case, the event belongs to one refresh target and it is for sure the refresh would create only one task.

lfu: Other than returning ```task_ids``` in an array, the refresh may also return [```nil```](https://github.com/ManageIQ/manageiq/blob/962af5d7a934d8d286ac2f938a5d5e05c31efd8f/app/models/ems_refresh.rb#L56) when target is empty.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/target_refresh.rb
```diff
@@ -0,0 +1,50 @@
...
+        module StateMachines
+          module Refresh
+            class TargetRefresh
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                task_id = event.refresh(refresh_target, true)
+                @handle.set_state_var(:refresh_task_id, task_id)
```
mkanoor: @lfu you mentioned that the task_id can be nil if there are no targets, should we be stopping the state machine with an error if there are no task ids instead of storing a nil

lfu:
Working on test cases for ```target_refresh.rb```.

lfu:
Will do an end-to-end testing after my evm server is up.

content/automate/ManageIQ/System/Event/StateMachines/Refresh.class/__methods__/check_refreshed.rb
```diff
@@ -0,0 +1,59 @@
...
+              def main
+                check_status(refresh_task)
+              end
+
+              private
+
+              def check_status(task)
+                case task.state
+                when STATE_FINISHED
+                  @handle.root['ae_result'] = task.status == STATUS_OK ? 'ok' : 'error'
```
mkanoor: @lfu 
If there is an error we should show it in this log. MiqTask#message would have the error message.

lfu:
End-to-end testing worked well.

gmcculloug:
@tinaafitz Can you look this over again as some things have changed since you approved.

lfu:
Change in https://github.com/ManageIQ/manageiq/pull/16868 caused the Travis failure.

mfalesni:
What is the best way to test this change? Where does the state machine get called so I can see how does it work?

lfu:
@mfalesni The state machine is called by [reconfigvm_task_complete](https://github.com/ManageIQ/manageiq-content/blob/e0ec80069f3a5e2c75079fa519e9dfbf18e6771d/content/automate/ManageIQ/System/Event/EmsEvent/EVM.class/reconfigvm_task_complete.yaml#L12) event.
Run Vm Reconfigure task via UI which would call the state machine at the end of the process. It would be even better if somehow the provider refresh takes longer time to get done.

simaishi:
Fine backport details:

```
$ git log -1
commit 473466ed60cf61e69ac7f1cad7ea9b5d08958cbd
Date:   Thu Jan 25 09:02:48 2018 -0500

    Merge pull request #243 from lfu/event_action_refresh_sync
    
    Add event state machine for refresh.
    (cherry picked from commit e0ec80069f3a5e2c75079fa519e9dfbf18e6771d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1550725
```

simaishi:

```
$ git log -1
commit 068779473d7cbc266754277d9b14793df00dc90c
Date:   Thu Jan 25 09:02:48 2018 -0500

    Merge pull request #243 from lfu/event_action_refresh_sync
    
    Add event state machine for refresh.
    (cherry picked from commit e0ec80069f3a5e2c75079fa519e9dfbf18e6771d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1550724
```

PR #248, opened 2018-02-08
V2V - State machines - Initial work

fdupont-redhat:
This is the initial work to integrate the state machines and methods required to transform virtual machines.
There are three main areas:
 - Transformation Host:
    - Enable/disable role (provision host)
    - Methods to assign and track transformation hosts in tasks
 - Transformation Request:
    - Approve the request, i.e. select the VM to be transformed and create tasks
    - Schedule tasks based on dependencies, capacity, throttling rules...
- Transformation Task:
    - Execute the transformation
    - Track completion and provide machine readable progress for UI


fdupont-redhat:
@miq-bot add_label v2v

gmcculloug:
@bzwei Please review

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/tagasdisabled.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: 1.0
+object:
+  attributes:
+    name: TagAsDisabled
+    display_name: 
+    description: 
+    scope: instance
+    language: ruby
+    location: inline
+    options: {}
```
mkanoor: @fdupont-redhat Is this missing the embedded methods data since your method is using the embedded methods

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/transformvm_vmware2redhat_vddk.yaml
```diff
@@ -0,0 +1,13 @@
...
+version: 1.0
+object:
+  attributes:
+    name: TransformVM_vmware2redhat_vddk
+    display_name: 
+    description: 
+    scope: instance
+    language: ruby
+    location: inline
+    options: {}
```
mkanoor: @fdupont-redhat This might also be missing the embedded_methods names.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -0,0 +1,66 @@
...
+          @debug = true
+          @handle = handle
+        end
+
+        def get_transformation_hosts(ems)
+          ems.hosts.select { |host| host.tagged_with('mf_transformation_host', 'true') }
+        end
+        
+        def get_runners_count_by_host(host)
+          return @handle.vmdb(:vm).all.select { |v| v.custom_get('Transformation Host') == host.name }.size
```
mkanoor: @fdupont-redhat dont need the return here

gmcculloug:
@fdupont-redhat It would be good if you could run through the rubocop warning/errors and address those as a first pass.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -0,0 +1,66 @@
...
+        end
+        
+        def get_runners_count_by_ems(ems)
```
mkanoor: @fdupont-redhat 
``` ruby
def get_runners_count_by_ems
  get_transformation_hosts.inject(0){|sum,host| sum + get_runners_count_by_host(host) }
end
```

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
mkanoor: @fdupont-redhat This can be an expensive call fetching all the vms and then fetching/comparing host name.
Its good in the sense that it doesn't require locking the host record to add/remove vms being transformed

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -0,0 +1,66 @@
...
+            if transformation_host.nil? || runners_count < min_runners
+              transformation_host = host
+              min_runners = runners_count
+            end
+          end
+          if transformation_host.nil?
+            @handle.log(:info, "No transformation host found.")
+          else
+            @handle.log(:info, "Least busy transformation host: #{transformation_host}")
+            vm.custom_set('Transformation Host', transformation_host.guid)
```
mkanoor: @fdupont-redhat are transformation_host.guid and transformation_host.name the same, since in line # 19 you are using name.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -0,0 +1,66 @@
...
+          end
+          return transformation_host
+        end
+        
+        def unset_transformation_host(vm)
+          vm.custom_set('Transformation Host', nil)
```
mkanoor: @fdupont-redhat can the custom attribute be deleted instead of being set to nil. Will the nil logic be used elsewhere to compare completed transformations.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -0,0 +1,66 @@
...
+          # Determine the least utilized conversion host
+          min_runners, transformation_host = nil, nil
+          
+          get_transformation_hosts.each do |host|
+            runners_count = get_runners_count_by_host(host)
+            max_runners = host.custom_get('Max Transformation Runners')
+            max_runners = factory_config['transformation_host_max_runners'] if max_runners.blank?
+            next if runners_count == max_runners
+            ( transformation_host = host ; break ) if runners_count == 0
+            if transformation_host.nil? || runners_count < min_runners
```
mkanoor: @fdupont-redhat How is the min_runners being set, it starts out as nil and only gets set inside of the if block to the runner count. The || logic seems odd.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
gmcculloug: This is how you do it.  There are only three methods exposed `custom_keys, custom_get` and `custom_set`.

See [/app/models/mixins/custom_attribute_mixin.rb#L82](https://github.com/ManageIQ/manageiq/blob/d4a91f5434110a5be2c2cbdddce40349626e6bf1/app/models/mixins/custom_attribute_mixin.rb#L82)

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
```diff
@@ -0,0 +1,15 @@
...
+object:
+  attributes:
+    display_name: 
+    name: Check
+    inherits: 
+    description: 
+  fields:
+  - State1:
+      value: "/Transformation/TransformationHost/${#transformation_host_type}/RoleCheck"
+      on_error: "/Transformation/TransformationHost/${#transformation_host_type}/MarkAsDisabled"
```
mkanoor: @fdupont-redhat 
We dont support paths in on_error/on_exit/on_entry slots?

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/roleenable.yaml
```diff
@@ -0,0 +1,154 @@
...
+    options: {}
+  inputs:
+  - field:
+      aetype: 
+      name: v2v_content_server
+      display_name: 
+      datatype: string
+      priority: 0
+      owner: 
+      default_value: bastion.v2v.example.com
```
mkanoor: @fdupont-redhat the default value should be empty

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/roleenable.yaml
```diff
@@ -0,0 +1,154 @@
...
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: 
+      name: v2v_vddk_package_url
+      display_name: 
+      datatype: string
+      priority: 1
+      owner: 
+      default_value: http://bastion.v2v.example.com/vddk/VMware-vix-disklib-6.5.2-6195444.x86_64.tar.gz
```
mkanoor: @fdupont-redhat please check default value

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
fdupont-redhat: If you have a less expensive approach, I take.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
fdupont-redhat: It will be changed, as the transformation host will be part of the `options` hash of the transformation task, rather that the VM.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
fdupont-redhat: It will be changed, as the transformation host will be part of the options hash of the transformation task, rather that the VM.

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
fdupont-redhat: Well, a typo while copying. It should be `/Transformation/TransformationHost/${#transformation_host_type}.MarkAsDisabled`.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
gmcculloug: @fdupont-redhat You could lookup the CustomAttribute records that has the matching `name` and `value` properties and get the count from there.  If needed you can limit the result set to just VMs by only returning ones with `resource_type  => VmOrTemplate`.

Maybe `@handle.vmdb(:custom_attribute).where(:name => 'Transformation Host', :value => host.name).size`

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/tagasdisabled.yaml
fdupont-redhat: Where do you see it using an embedded method ? That one doesn't. Anyway, I see I forgot to rename the module from ConversionHost to TransformationHost.

gmcculloug:
@fdupont-redhat Can you cleanup the two `Style/YodaCondition` rubocop issues and un-`wip` if this is ready?

Not sure why we are seeing the other rubocop about parameters at the moment.

fdupont-redhat:
@gmcculloug These errors are strange, as this syntax is found in all the methods that implement the module layout. For example: https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Cloud/Orchestration/Operations/Methods.class/__methods__/available_tenants.rb.

gmcculloug:
Correct, the `YodaCondition` is a more recent rubocop check so we were not seeing that when the new method type was introduced.

Ultimately we will want to go through and correct them all, but the goal is not to introduce more warnings.

fdupont-redhat:
@gmcculloug WIP label removed and all lights green on rubocop side.

gmcculloug:
@tinaafitz @mkanoor Can you give this another look after changes?

fdupont-redhat:
Forgot to add the request entrypoints for the role check/enable/disable state machines.
@miq-bot add_label wip

@tinaafitz @lfu What would you advise for the request names ?

content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/tagasdisabled.yaml
mkanoor: @fdupont-redhat in your earlier ruby code you had

```
if @debug
      ManageIQ::Automate::System::CommonMethods::MIQ_AE::Debug.new.dump_object
      ManageIQ::Automate::System::CommonMethods::MIQ_AE::Debug.new.dump_root
end
``` 
which would have needed the embedded methods, now that you have removed it, its not relevant.

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
```diff
@@ -0,0 +1,15 @@
...
+object:
+  attributes:
+    display_name: 
+    name: Check
+    inherits: 
+    description: 
+  fields:
+  - State1:
+      value: "/Transformation/TransformationHosts/${#transformation_host_type}/RoleCheck"
+      on_error: "/Transformation/TransformationHosts/${#transformation_host_type}.MarkAsDisabled"
```
lfu: Where is ```MarkAsDisabled``` defined?

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - State2:
+      value: "/Transformation/TransformationHosts/${#transformation_host_type}.MarkAsEnabled"
```
lfu: Same here.

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
mkanoor: There is a TagAsEnabled? Does this needs to be renamed.

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
mkanoor: @fdupont-redhat Does this need a METHOD:: since its a call to a class method

mkanoor: @fdupont-redhat Also is this TagAsEnabled?

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
fdupont-redhat: You're right. It is TagAsDisabled. I have renamed it.

content/automate/ManageIQ/Transformation/TransformationHosts/StateMachines/TransformationHostRole.class/check.yaml
fdupont-redhat: You're right. It is TagAsDisabled. I have renamed it.

@mkanoor `METHOD::` is not needed because `on_error` expects method(s). I don't know how it is coded internally, but experimentally it works without it. If you think I should make it explicit, I'll change it.

fdupont-redhat:
@miq-bot remove_label wip

simaishi:

```
$ git log -1
commit cfa1f77759fdd22e75ff7e19a770cb6d9bd0d0da
Date:   Thu Feb 15 15:55:17 2018 -0500

    Merge pull request #248 from fdupont-redhat/v2v_transformation_state_machines_init
    
    V2V - State machines - Initial work
    (cherry picked from commit 15b3a0e8e859ced52cd66c6ff0ecce91cf47a581)
```

PR #250, opened 2018-02-15
[GAPRINDASHVILI] Support symlinked spec directories (for testing client Automate domains)

jeffwarnica:
In working on testing some custom automate code, I noticed that the rake tasks were not picking up on symlinks into _spec.rb locations. This seems to allow that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /spec/manageiq
 /spec/reports/
 /tmp/
+/.idea/
--- a/README.md
+++ b/README.md
@@ -34,32 +34,29 @@ Now you are ready to begin development.  You can run the specs with
 
 Please be sure to add specs for any new automate methods you create, and follow
 the [ManageIQ development guidelines](https://github.com/ManageIQ/guides/blob/master/coding_style_and_standards.md).
-Thanks for your contribution!
+
 
 ## Testing your own Automate Domain
 
 It is possible to test your own ManageIQ Automate Domain methods leveraging this framework.
 
-Individual Automate methods need to be converted to the [Class Style](https://github.com/ManageIQ/manageiq-content/issues/8), and
-tests written.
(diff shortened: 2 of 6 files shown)


jeffwarnica:
If you can help with the gitfu, I can rebase.

I have no strong opinions on my developed layout, it seems consistent. If there are no symlinks to something else, then nothing will happen. If there is a more conventional way to do this, fine by me.

11:12 $ find -type l
./content/automate/Cflab
./spec/content/automate/Cflab
./spec/factories/Cflab
./spec/manageiq
✔ ~/Projects/manageiq-content [gaprindashvili|✚ 1…5] 
11:12 $ find -type l|xargs readlink -f
/home/jwarnica/Projects/manageiq-content-sample-tests/automate/Cflab
/home/jwarnica/Projects/manageiq-content-sample-tests/spec/Cflab
/home/jwarnica/Projects/manageiq-content-sample-tests/factories/Cflab
/home/jwarnica/Projects/manageiq

targeting the brand new https://github.com/jeffwarnica/manageiq-content-sample-tests

gmcculloug:
@jeffwarnica If you ultimate goal here is to test your own repo methods we need to work toward adding your repo to the `Gemfile` and registering it as an automate domain so it gets loaded when the Automate datastore is reset.

We should not have to modify this repo to support adding new ones.

Let me know and we can help on getting the proper repo configuration in place.  Would be good to capture the steps for others the reference in the future.

jeffwarnica:
@gmcculloug As is today, my automate methods are not run from the datastore (that is, they are run from the file system). Adding them to the datastore seems contrary to the desire of fast tests.

My goal is not to facilitate domain/s entering the build pipeline, but to simply unit test them. I'd never want to add my code to manageiq-content. Real world automate code would come from some other repository... and be linked in.

A mechanism for adding, say, the RedHat domain, into the build pipeline is something else.

It is an ugly hack all around to run automate code outside of the sandbox... and this works. I see the PR is a bit messy (rebase fail), but what is essentially a 2 line code change seems minimal.

PR #251, opened 2018-02-20
[WIP] Spec testing for embedded and caller method

mkanoor:
This is a sample for embedded method its spec and the caller method and its spec.

The embedded method is written as a class and it can be tested independently.

The method that uses the embedded methods can be tested by stubbing out the instances and just testing the method without testing the embedded method.


mkanoor:
@tinaafitz @billfitzgerald0120 @gmcculloug 
Please review

spec/content/automate/ManageIQ/System/CommonMethods/StateMachineMethods.class/__methods__/utility_spec.rb
```diff
@@ -0,0 +1,22 @@
+require_domain_file
+
+describe ManageIQ::Automate::System::CommonMethods::StateMachineMethods::Utility do
+  context "class method" do
+    it "convert. in name to underscore" do
```
tinaafitz: @mkanoor Can you add a space before the . here?

PR #252, opened 2018-02-22
Add VM Transformation state machine

fdupont-redhat:
This state machine describes the states to transform a virtual machine. It uses nested state machines to easily change the behavior based on source and destination EMS, as well as transformation method.

The `update_status.rb` method also add tracking information, such as start and update times, and percentage of progress. The percentage is either provided by the method itself, or defaults to the number of retries on the max retries of the state.

It also take into account the nesting by creating an artificial ancestry. The ancestry requires the state machine designer to add the following argument to the URI: `?state_ancestry=${#state_ancestry}/${#ae_state}"`.


gmcculloug:
cc @bzwei

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
```diff
@@ -0,0 +1,119 @@
...
+            state_hash = progress['states']["#{state_ancestry}/#{state_name}"]
+
+            # Get the state progress data provided by the state method
+            state_progress = @handle.get_state_var('ae_state_progress')
+            
+            # Aggregate the state progress in the state hash based on state
+            # phase (entry, exit, error). This updates the details with live
+            # data from the method, and also sets sane defaults.
+            case @handle.root['ae_status_state']
+            when 'on_entry'
```
gmcculloug: This `main` method should be broken down into smaller methods.  A good place to start is converting these `when` blocks for the case statement into separate methods.

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
```diff
@@ -0,0 +1,119 @@
...
+      class UpdateStatus
+        def initialize(handle = $evm)
+          @handle = handle
+        end
+
+        # This method gets the current state, even in the context of nested
+        # state machines. Where running a nested state machines, there can be
+        # more than one active states: one per level of nesting. The real
+        # active state is the one with the deepest ancestry.
+        def current_state(states)
```
mkanoor: @fdupont-redhat 
Couldn't you get the current state from root['ae_state']?
We are setting it here https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_state_machine.rb#L63

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
```diff
@@ -0,0 +1,119 @@
+module Transformation
```
mkanoor: @fdupont-redhat To stay consistent with other ManageIQ methods please add ManageIQ and Automate
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb#L8

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
fdupont-redhat: In the context of nested state machines, `root [ae_state]` contains the name of the current state in the current state machine. And nothing prevents from having two states with the same name when using nested state machines. So, here I take the state ancestry depth as a criteria to distinguish the real running state.

mkanoor: @fdupont-redhat 
So do you need the fully qualified instance name that is being used at that moment? We have that information we can set it as an extra attribute in the root object.

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
```diff
@@ -0,0 +1,119 @@
...
+            if @handle.root['ae_status_state'] == 'on_entry'
+              progress['current_state'] = current_state(progress['states'])
+              progress['current_description'] = progress['states'][progress['current_state']]['description']
+            # End if we exit/error the state, we update the percentage based on
+            # children states and current state weight.
+            else
+              progress['percent'] = reconcile_children_percent('', progress['states'])
+            end
+            # We record the progress as a task option.
+            task.set_option(:progress, progress)
```
bzwei: we have new method `task.update_transformation_progress(progress)`

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__methods__/update_status.rb
```diff
@@ -0,0 +1,119 @@
...
+              # set percentage to 100 and status to 'finished'. We also merge
+              # the potential message from method.
+              else
+                state_hash['status'] = 'finished'
+                state_hash['percent'] = 100
+                state_hash['message'] = state_progress.nil? ? "#{state_name} is finished, moving on." : state_progress['message']
+              end
+              # Then, we set the update time to now, which is also finish time
+              # when the state is finished.
+              state_hash['updated_on'] = Time.now.utc
```
bzwei: The task itself has `updated_on` column. Should not need this one.

michalskrivanek:
I guess you're renaming "rhev" to either ovirt or rhv:)
And the steps which are currently designed to work with export domain would be simplified with the upload api - no sysprep, that disk preparation steps...those should be all not required

gmcculloug:
@mkanoor Can you give this one last review?

simaishi:

```
$ git log -1
commit fd65978a1876d656e176729b2d2d67ee3057261e
Date:   Wed Mar 21 17:26:20 2018 -0400

    Merge pull request #252 from fdupont-redhat/v2v_state_machines_vm_transformation
    
    Add VM Transformation state machine
    (cherry picked from commit d45d46f2f2801eeba43d16dae2a7c9c8482fa717)
```

PR #253, opened 2018-02-28
Parse generic multi-select dialog elements

branic:
This adds support for parsing generic service dialog elements that have multi-select option enabled. This also fixes parsing option service dialog elements that are multi-select and not a VMDB object.

This PR address https://bugzilla.redhat.com/show_bug.cgi?id=1459344

I think this PR should be back ported to Fine and Gaprindashvili.

### DIFF
--- a/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
+++ b/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
@@ -18,9 +18,13 @@ def process_comma_separated_object_array(sequence_id, option_key, value, hash)
   return if value.nil?
   options_value_array = []
   value.split(",").each do |entry|
+    next if entry.blank?
     vmdb_obj = vmdb_object_from_array_entry(entry)
-    next if vmdb_obj.nil?
-    options_value_array << (vmdb_obj.respond_to?(:name) ? vmdb_obj.name : "#{vmdb_obj.class.name}::#{vmdb_obj.id}")
+    options_value_array << if vmdb_obj.nil?
+                             entry
+                           else
+                             (vmdb_obj.respond_to?(:name) ? vmdb_obj.name : "#{vmdb_obj.class.name}::#{vmdb_obj.id}")
+                           end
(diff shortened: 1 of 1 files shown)


content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
```diff
@@ -18,9 +18,13 @@ def process_comma_separated_object_array(sequence_id, option_key, value, hash)
...
   options_value_array = []
   value.split(",").each do |entry|
+    next if entry.blank?
     vmdb_obj = vmdb_object_from_array_entry(entry)
-    next if vmdb_obj.nil?
-    options_value_array << (vmdb_obj.respond_to?(:name) ? vmdb_obj.name : "#{vmdb_obj.class.name}::#{vmdb_obj.id}")
+    options_value_array << if vmdb_obj.nil?
+                             entry
+                           else
+                             (vmdb_obj.respond_to?(:name) ? vmdb_obj.name : "#{vmdb_obj.class.name}::#{vmdb_obj.id}")
```
mkanoor: @branic Can you explain your use case for the vmdb objects, its getting converted to names from ids. Wouldn't ID's have been more useful? From discussions with @tinaafitz it seems like we needed names for tags but for other vmdb objects ID would have been a better choice than names. I am ok with leaving it as names as long as it fulfills your use case.

branic: @mkanoor There are cases where a multi-select Service Dialog element is not based on VMDB objects. For example a Service Dialog drop-down element can be created with multi-select enabled. The name/value pairs added to the element will not be VMDB objects (and therefore not have IDs), but the selected values are needed by the automation code. Without this change the dialog_parser doesn't parse those multi-select elements and add as such they won't be added to the tasks that are created for the Service Catalog Item that is ordered.

This PR retains current functionality where a multi-select element that is a VMDB object (such as from a Tag Control) will have the VMDB objects added instead of the name string .

I can provide an example if that would help to understand the use case.

PR #256, opened 2018-03-02
[FINE] Remove the options hash which is new in version 5.9.

lfu:
Depends on https://github.com/ManageIQ/manageiq/pull/17081.
Follow up of https://github.com/ManageIQ/manageiq-content/pull/243.

https://bugzilla.redhat.com/show_bug.cgi?id=1550725

@miq-bot add_label bug, fine/yes


gmcculloug:
@lfu Tests are failing because there is no existing service model for MiqTask.  In `Gaprindashvili` and master this service model would be auto-generated so it does not fail there.

PR #259, opened 2018-03-07
Refactoring cloud vm retirement pre_retirement method

pkomanek:
Purpose or Intent
---------------------

Refactoring Cloud/VM/Retirement/StateMachines/Methods.class/__methods__/pre_retirement.rb method with spec. This PR is based on the issue bellow.

Links
-------------------
#8


content/automate/ManageIQ/Cloud/VM/Retirement/StateMachines/Methods.class/__methods__/pre_retirement.rb
```diff
@@ -1,10 +1,35 @@
...
 
-vm = $evm.root['vm']
-if vm && vm.power_state == 'on'
-  ems = vm.ext_management_system
-  $evm.log('info', "Stopping Instance <#{vm.name}> in EMS <#{ems.try(:name)}>")
-  vm.stop if ems
+                def main
+                  vm = @handle.root['vm']
+                  if vm && vm.power_state == 'on'
+                    ems = vm.ext_management_system
```
mkanoor: @tinaafitz if the ems is missing and the vm is powered on, should we be raising an exception and stopping the workflow or should we keep continuing. I remember we had some corner cases that we were trying to solve where a failed retirement is restarted. If there is no ems should we log the fact that its an orphaned vm.

mkanoor:
@pkomanek 
When you run the specs can you check the code coverage for the method
from the shell set an environment variable CI, run the spec and check coverage for the file

```
export CI=1
bundle exec rspec <spec_file_name>
In the current directory check for a file under coverage/index.html and see if all lines have been tested. 
```

spec/content/automate/ManageIQ/Cloud/VM/Retirement/StateMachines/Methods.class/__methods__/pre_retirement_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+  end
+
+  it 'powers off a vm in a \'powered_on\' state' do
+    expect(svc_vm).to receive(:stop)
+    described_class.new(ae_service).main
+  end
+
+  it 'does not stop a vm in \'powered_off\' state' do
+    vm.update_attribute(:raw_power_state, "PowerOff")
```
gmcculloug: Use: `vm.update_attributes(:raw_power_state => "PowerOff")`

content/automate/ManageIQ/Cloud/VM/Retirement/StateMachines/Methods.class/__methods__/pre_retirement.rb
tinaafitz: @mkanoor We should continue processing if the ems is nil and the VM will have the power state of "unknown" in that case

content/automate/ManageIQ/Cloud/VM/Retirement/StateMachines/Methods.class/__methods__/pre_retirement.rb
```diff
@@ -1,10 +1,35 @@
...
+                    @handle.log('info', "Stopping Instance <#{vm.name}> in EMS <#{ems.try(:name)}>")
```
tinaafitz: @pkomanek Can you issue the log message only if there is an ems? You can remove the try from the log message in that case.
something like 
if ems
  @handle.log('info', "Stopping Instance <#{vm.name}> in EMS <#{ems.name}>")
  vm.stop
end

tinaafitz:
bump @mkanoor @gmcculloug

PR #261, opened 2018-03-13
Add Azure events for targeted refresh

tumido:
Provide automated refresh on events which are recognized by `event_target_parser` in Azure.

Related to: https://github.com/ManageIQ/manageiq-providers-azure/pull/222

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Azure.class/images_delete_endrequest.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Azure.class/images_delete_endrequest.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: images_delete_EndRequest
+    inherits: 
+    description: 
+  fields:
+  - rel4:
+      value: "/System/event_handlers/refresh"
--- a/content/automate/ManageIQ/System/Event/EmsEvent/Azure.class/loadbalancers_delete_endrequest.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/Azure.class/loadbalancers_delete_endrequest.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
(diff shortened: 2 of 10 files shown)


gmcculloug:
@lfu Please review

gmcculloug:
@tumido Any reason to keep the `wip` label?

tumido:
Just checking if the events get propagated correctly.

We have some events which are inconsistent in names e.g.: `publicIPAddresses_delete_EndRequest` and `publicIpAddresses_delete_EndRequest`. So I'm checking if it has any impact. Maybe you can help me? :wink:

Does the `name` field in the YAML has any importance? I know the proper YAML file is located based on downcased matching against `event_type`, so we should be fine at this point. However, I'm uncertain if it can or cannot cause any harm in any step after that...

We can't have multiple files for the same event, which differs in case-sensitive matching, however downcased are the same. (I know this is our mess and we have to clean it, but we can't do so at this point... Because: reasons, lol.)

Ladas:
This is actually Azure's mess, they don't keep consistent resource type across events (event though it's documented correctly). But I am almost sure we pick the handle by the down-cased string, right @gmcculloug @lfu ?

tumido:
So, I've experimentally confirmed, the events are picked up correctly and the refresh lands fine in the `MiqQueue`. Removing wip. Thanks for the patience. We want to be sure what we're doing, before we break things, lol. :four_leaf_clover: 

@miq-bot remove_label wip

gmcculloug:
@tumido Thanks for testing that out.

tumido:
@miq-bot add_label gaprindashvili/yes

simaishi:

```
$ git log -1
commit eee3cc8ece367ff2b878f8ebb6b00d7fec151608
Date:   Thu Mar 15 14:36:09 2018 -0400

    Merge pull request #261 from tumido/azure_events
    
    Add Azure events for targeted refresh
    (cherry picked from commit af63ed2d5c8942985e4389e4960ae0ac3a22c01b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1558078
```

PR #262, opened 2018-03-13
Add task status update to retirement status

d-m-u:
The service retirement status should have the status included in the message since retirement is morphing into a request. 

## Related to: 
classic-ui pr here: https://github.com/ManageIQ/manageiq-ui-classic/pull/3409
https://github.com/ManageIQ/manageiq-automation_engine/pull/157

## Depends on:
https://github.com/ManageIQ/manageiq/pull/16933


d-m-u:
@tinaafitz I think this works?

d-m-u:
@mkanoor this could get reviewed too! thanks!

mkanoor:
@d-m-u 
Can you update the spec file for this.
If you run with the following commands 


export CI=1
rspec .../your_spec_file

open coverage/index.html

You would be able to see if you have coverage for all lines of code

d-m-u:
@mkanoor I think it's just the generic status update for services, the content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/update_status.rb, right?

d-m-u:
screen shot 2018-03-26 at 2 33 19 pm

content/automate/ManageIQ/Service/Retirement/StateMachines/ServiceRetirement.class/__methods__/update_service_retirement_status.rb
```diff
@@ -6,6 +6,7 @@
 server = $evm.root['miq_server']
```
tinaafitz: @d-m-u This is the correct file name, so you'll want to copy your changes from update_serviceretirement_status.rb in this PR to this file and remove the  update_serviceretirement_status.rb file from this PR. ...ment/StateMachines/ServiceRetirement.class/__methods__/update_serviceretirement_status.rb

content/automate/ManageIQ/Service/Retirement/StateMachines/ServiceRetirement.class/__methods__/update_serviceretirement_status.rb
```diff
@@ -0,0 +1,58 @@
...
+            class UpdateServiceRetirementStatus
+
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                task = @handle.root['service_retire_task']
+
+                if task.nil?
```
tinaafitz: @d-m-u Normally, this would be correct in that we would raise if an object we expected to be present was missing. In this case though, we want to be backward compatible, so you can remove the the if task.nil? code block.

tinaafitz:
@mkanoor Please review.

PR #264, opened 2018-03-22
Modify on_{entry,exit,error} path to use WeightedUpdateStatus

fdupont-redhat:
Realized the state machine still used update_status, which doesn't exist.
This PR fixes it, so that it used `/System/CommonMethods/MiqAe.WeightedUpdateStatus`


fdupont-redhat:
@miq-bot add_label transformation

PR #273, opened 2018-04-05
[WIP] Changed Cloud VM Provision validate_request max_memory to be consistent.

billfitzgerald0120:
*** Important Note - This will change the existing cloud request approval behavior and, as a result, should NOT be backported.

The values for approval are confusing as Jeffrey Cutter mentioned in his BZ.

I changed instance values to be in Megabytes.  They were in bytes.
The instance values for quota are in megabytes so the approval values will match now.

image

The value 1024 was 1 Megabyte, with this change it will be 1 Gigabyte.

Tag values should match the UI and be in gigabytes.  They were in Megabytes.
I changed tag values to be in Gigabytes.

image

If you set the 1GB tag value on a template, it was 1MB.  Now it will be 1GB.

I also human_sized all the log messages so going forward it will be clear what the values really are.

INFO -- : <AEMethod validate_request> Auto-Approval Request was not auto-approved for the following reasons: (Requested Memory 2 GB limit is 1 GB)

I will be doing another PR for Infrastructure with the same changes.

https://bugzilla.redhat.com/show_bug.cgi?id=1553530

### DIFF
--- a/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
+++ b/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
@@ -126,7 +126,8 @@ max_memory = nil
 # Use value from model unless specified above
 max_memory ||= $evm.object['max_memory']
 unless max_memory.nil?
-  $evm.log("info", "Auto-Approval Threshold(Model):<max_memory=#{max_memory}> detected")
+  max_memory = max_memory.to_i.megabytes
+  $evm.log("info", "Auto-Approval Threshold(Model):<max_memory=#{max_memory.to_s(:human_size)}> detected")
 end
 
 # Reset to nil if value is zero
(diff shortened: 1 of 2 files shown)


billfitzgerald0120:
@miq-bot add_label bug, gaprindashvili/yes

billfitzgerald0120:
@miq-bot assign @gmcculloug

gmcculloug:
@billfitzgerald0120 Can you explain this PR a little more with respect to how it might impact existing implementations?  Can this be safely back-ported?

I think some examples of exactly what is begin changed would be helpful in the description.

billfitzgerald0120:
@gmcculloug I updated the description.  The tag values are wrong and should be changed. It says gigabytes when its really doing megabytes.  The instance value doesn't really say what the unit is. The impact would be, If existing customers are using a memory value for approval, then some requests that weren't auto-approved before might be approved with these changes.

JPrause:
@billfitzgerald0120 can you add the fine/yes label if this can be backported to that branch.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
```diff
@@ -126,7 +126,8 @@ def flavor_value(prov_resource, option)
 # Use value from model unless specified above
 max_memory ||= $evm.object['max_memory']
```
syncrou: @billfitzgerald0120 - is it always the case that `$evm.object['max_memory']` comes in the same way? The reporter on the BZ made mention that Amazon does one thing ( bytes ) but what about other cloud providers?

tinaafitz: @syncrou Yes, ```$evm.object['max_memory']``` always comes in the same way. The provider values differ, but that doesn't affect this instance variable.

tinaafitz:
@miq-bot remove_label gaprindashvili/yes

tinaafitz:
I spoke to Brad about this change and it was agreed that it would NOT be backported.

tinaafitz:
@syncrou Please review.

PR #274, opened 2018-04-06
[WIP] Changed Infra VM Provision validate_request max_memory to be consistent.

billfitzgerald0120:
*** Important Note - This will change the existing infrastructure request approval behavior and, as a result, should NOT be backported.

See similar changes for cloud PR: https://github.com/ManageIQ/manageiq-content/pull/273
Added tests for VMware.

The values for approval are confusing as Jeffrey Cutter mentioned in his BZ.

I changed instance values to be in Megabytes. They were in bytes.
The instance values for quota are in megabytes so the approval values will match now.

Tag values should match the UI and be in gigabytes. They were in Megabytes.
I changed tag values to be in Gigabytes.

I also human_sized all the log messages so going forward it will be clear what the values really are.

https://bugzilla.redhat.com/show_bug.cgi?id=1553530

### DIFF
--- a/content/automate/ManageIQ/Infrastructure/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
+++ b/content/automate/ManageIQ/Infrastructure/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
@@ -113,7 +113,8 @@ max_memory = nil
 # Use value from model unless specified above
 max_memory ||= $evm.object['max_memory']
 unless max_memory.nil?
-  $evm.log("info", "Auto-Approval Threshold(Model):<max_memory=#{max_memory}> detected")
+  max_memory = max_memory.to_i.megabytes
+  $evm.log("info", "Auto-Approval Threshold(Model):<max_memory=#{max_memory.to_s(:human_size)}> detected")
 end
 
 # Reset to nil if value is zero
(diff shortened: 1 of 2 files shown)


JPrause:
@billfitzgerald0120 can you add the fine/yes label if this can be backported to that branch.

content/automate/ManageIQ/Infrastructure/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
```diff
@@ -113,7 +113,8 @@
 # Use value from model unless specified above
 max_memory ||= $evm.object['max_memory']
 unless max_memory.nil?
-  $evm.log("info", "Auto-Approval Threshold(Model):<max_memory=#{max_memory}> detected")
+  max_memory = max_memory.to_i.megabytes
```
syncrou: @billfitzgerald0120 - Same question as [here](https://github.com/ManageIQ/manageiq-content/pull/273/files#r202179618) if we can always trust that all infrastructure providers use the same base numbers for `max_memory`

tinaafitz:
@miq-bot remove_label gaprindashvili/yes

tinaafitz:
Removed the gaprindashvilli label since this should NOT be backported.

content/automate/ManageIQ/Infrastructure/VM/Provisioning/StateMachines/ProvisionRequestApproval.class/__methods__/validate_request.rb
tinaafitz: @syncrou The value will be consistent here. The provider values could differ in the base numbers for the requested values.

tinaafitz:
@syncrou Please review.

spec/automation/unit/method_validation/vmware_validate_request_spec.rb
```diff
@@ -0,0 +1,69 @@
+describe "Auto Approval Request Validation for Infrastructure" do
+  include Spec::Support::QuotaHelper
+  let(:ws) { MiqAeEngine.instantiate("/System/Request/Call_Method?#{method}&#{args}&#{@value}", @user) }
```
syncrou: I really like these tests, but I wish I had an understanding of the underlying values ( not changed since not in the diff ) but I'm finding the error messages are only one sided and I'm not sure what they're proving.

PR #283, opened 2018-04-19
Move the task message into check for task presence

d-m-u:
The updated message should only include the task message if the task exists.


d-m-u:
@tinaafitz please review
@miq-bot add_label bug

PR #300, opened 2018-05-22
Updated Configuration/Email class and existing V2v instances.

billfitzgerald0120:
Modified class schema to take advantage of accessing object information from the Configuration/Email namespace.
Modified 7 existing V2v instances to use new class changes
and to use the new show_url.


content/automate/ManageIQ/System/Notification/Email.class/__class__.yaml
```diff
@@ -11,12 +11,52 @@ object:
...
+      datatype: boolean
+      priority: 1
+      owner: 
+      default_value: 'true'
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description:
```
mkanoor: @billfitzgerald0120 
Should we put a description that reads something like
"By default the email is enabled, to disable a specific email set this attribute to false"

PR #327, opened 2018-06-13
Set insecure_connection to target provider as default behavior.

fdupont-redhat:
During tests of V2V solution, we identified that it is highly probable that the certificate presented by RHV-M is not valid. virt-v2v-wrapper has an option to work in insecure mode and this PR makes it the default behavior to avoid transformation to fail due to an invalid certificate.

### DIFF
--- a/content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/vmtransform_vmwarews2rhevm_vddk.rb
+++ b/content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/vmtransform_vmwarews2rhevm_vddk.rb
@@ -51,18 +51,19 @@ module ManageIQ
                 virtv2v_networks = task[:options][:virtv2v_networks]
 
                 wrapper_options = {
-                  :vm_name            => source_vm.name,
-                  :transport_method   => 'vddk',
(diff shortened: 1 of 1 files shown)


fdupont-redhat:
@mkanoor It would mean changing the class schema. And this adds more complexity. Anyway, you're right it has to be configurable. I was going to create an issue to track the _v2v configuration_ topic as it is broader than certificates validation. But we don't know yet what component will be targeted to store configuration.

mkanoor:
@fdupont-redhat
I was thinking of adding a default value in the methods input parameter section and have it default to 
secure_connection => false.
In the ruby method you would be using @handle.inputs['secure_connection'] || true

Its your choice, I am ok with just updating the method with a Git Issue pointing to that line in the method to make it configurable

fdupont-redhat:
@mkanoor 
There's more than one way to do it, and that approach is elegant. Anyway, it's a broader topic about v2v configuration as a whole. At the moment, I'd rather have it hard coded and open an issue for 5.9.4.

simaishi:

```
$ git log -1
commit 752d25cab22c8e0521f92f596bc9c54ca5d19b41
Date:   Wed Jun 13 11:43:18 2018 -0400

    Merge pull request #327 from fdupont-redhat/v2v_insecure_connection_as_default
    
    Set insecure_connection to target provider as default behavior.
    (cherry picked from commit cd8a56f09fc22a882c89abafbd4f7589409184af)
```

PR #348, opened 2018-06-27
Add a common class to implement advisory lock

fdupont-redhat:
This PR aims at providing a generic lock mechanism for automate code. The concept is explained here: https://hashrocket.com/blog/posts/advisory-locks-in-postgres and an implementation is detailed here: https://gist.github.com/jeffwarnica/3fa2b94195d64b11ccc083155245133f. The idea here is to make it available easily in a method/class that can be easily embedded in other methods.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/advisorylock.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/advisorylock.rb
@@ -1,0 +1,40 @@
+module ManageIQ
+  module Automate
+    module System
+      module CommonMethods
+        module MiqAe
+          class AdvisoryLock
+            def initialize(handle = $evm)
+              @handle = handle
+            end
+
+            def main
+            end
+
+            def acquire_lock(lock_name)
+              hashcode = stable_hashcode(lock_name)
+              result = connection.select_value("select pg_try_advisory_lock(#{hashcode});")
+              (result == 't' || result == true)
+            end
+
+            def release_lock(lock_name)
+              hashcode = stable_hashcode(lock_name)
+              connection.execute("select pg_advisory_unlock(#{hashcode});")
+            end
+
+            private
+
+            def stable_hashcode(lock_name)
+              require 'zlib'
+              Zlib.crc32(lock_name) & 0x7fffffff
+            end
+          end
+        end
+      end
+    end
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
+  ManageIQ::Automate::System::CommonMethods::MiqAe::AdvisoryLock.new.main
+end
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/advisorylock.yaml
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/advisorylock.yaml
@@ -1,0 +1,13 @@
+---
+object_type: method
+version: 1.0
+object:
+  attributes:
+    name: AdvisoryLock
+    display_name: 
+    description: 
+    scope: instance
+    language: ruby
+    location: inline
+    options: {}
+  inputs: []


fdupont-redhat:

fdupont-redhat:

mkanoor:
@jrafanie @Fryguy 
Do you think its a good idea to have Postgres Database Advisory locks being created from Automate Methods?
or Would you rather prefer the back end model with/without an async mechanism do the locking and creating the valid resource and sending it back to the caller.

We have rejected many of these approaches in the past.

Fryguy:
> Do you think its a good idea to have Postgres Database Advisory locks being created from Automate Methods?

No.  I think my greatest concern here is that this is done in the automate *client* code, and not in the automation engine.  We generally restrict those, because of how dangerous they can be, and even thinking forward to the future when we might run Ruby code in containers, they just won't have access to the database at all, instead forcing them to go through the REST API.

> Would you rather prefer the back end model with/without an async mechanism do the locking and creating the valid resource and sending it back to the caller.
> 
> We have rejected many of these approaches in the past.

I did some searching, and this was tried in the past, but it was rejected.  I'm not *completely* against the idea, but @fdupont-redhat can you read over these previous attempts and show that the concerns raised there are not here.

Here is the original discussion: http://talk.manageiq.org/t/automate-engine-region-level-mutex/1185
and this is really 1 "PR" in chronological order, but the author chose to close and open new PRs:
https://github.com/ManageIQ/manageiq/pull/6577
https://github.com/ManageIQ/manageiq/pull/6680
https://github.com/ManageIQ/manageiq/pull/7047

jrafanie:
I see the value in a lock system for things such as integration with IPAM systems.  With that said, I don't know how we support x number of 800 line user scripts that leverage the lock system and want help debugging it.

* Is it possible to hold a lock too long and slow things down? (how do we even know this is what is happening).  I'm guessing we request the next ip address if we fail to get a lock on one that is sticking around too long.
* Can the locks build up if the client holds it too long (IPAM system is slow or the client code doesn't release the lock)?

How do we vet the code that is run around the locks? The linked gist makes use of object_send and instance_eval, which may get the job done, but aren't really the safest things to be doing.

So, while I think the feature is worthwhile, I don't know how we support/enhance it when the client code is so open to doing whatever it wants around locks.

What tooling can help us troubleshoot this when things aren't working?

fdupont-redhat:
About the fact that the lock may not be released or we may have deadlock, it will all be the responsibility of the Automate developer. Who blames Linux kernel developer for providing mutex mechanisms? The fact that it is provided doesn't mean that everybody will be using it for every use case. And today, it has been used by customers, with the help of consultants, so it seems beneficial.

The fact that we won't have access to the database once in containers is my only concern here and the only reason why I would do things differently. I propose this PR only because we don't have this kind of mechanism in Automate Engine. So, if someone knows how to implement it in AE, please do it or provide guidance :)

To answer @mkanoor questions from http://talk.manageiq.org/t/automate-engine-region-level-mutex/1185, here is a example method that I expect to use the advisory lock:

```
module Domain
  module Namespace
    module Class
      class Method
        ADVISORY_LOCK_NAME = 'my_unique_lock_name'

        def initialize(handle = $evm)
          @handle = handle
        end

        def exit_and_retry(message, interval='1.minutes')
          @handle.log(:info, "#{message}. Retrying.")
          @handle.root['ae_retry_interval'] = interval
        end
          
        def main
          advisory_locker = ManageIQ::Automate::System::CommonMethods::MiqAe::AdvisoryLock.new
          if advisory_locker.acquire_lock(ADVISORY_LOCK_NAME)
            # DO MY STUFF
          else
            exit_and_retry("Could not acquire lock", '10.seconds')
          end
        rescue => e
          @handle.set_state_var(:ae_state_progress, 'message' => e.message)
          raise
        ensure
          advisory_locker.release_lock(ADVISORY_LOCK_NAME)
        end
      end
    end
  end
end

if $PROGRAM_NAME == __FILE__
  Domain::Namespace::Class::Method.new.main
end
```

If we provide guidelines on how to use the mechanism, it can help solve problems. For example, here I use the implicit `begin` of `main` and add an `ensure` block to release the lock whatever happens.

mkanoor:
@fdupont-redhat 
After you have locked and you are running 
# DO My STuff
What prevents someone from running a $evm.instantiate or call another instance/method comes back .
How would you detect that circular reference?
Or create some other new lock, and they wait for each other to end.
These locks are running in external processes and when the process gets killed you might not get a chance to run the ensure code all the time.

These are distributed locks, which can span machines.
On the surface its looks trivial to implement a distributed lock but debugging these in customer environments is very difficult. And we have had issues with Automate methods doing this sort of thing in customer environments.

fdupont-redhat:
@mkanoor 
I know it's difficult to debug in customer environment, but https://gist.github.com/jeffwarnica/3fa2b94195d64b11ccc083155245133f shows that it is used. But I understand that there is a supportability issue here, because it will be shipped in the product.

However, for v2v, we need a way to solve race conditions in conversion host selection. On Gitter, the advisory lock was mentioned by @gmcculloug, and you already said it was a more complex problem. To be honest, I don't care how we implement it, but we need something. And @jeffwarnica and @dtrieu80 will also appreciate an official solution :wink:

mkanoor:
@fdupont-redhat 
For your specific problem with Transformation Host the solution might be in the app/models/ adding some logic and we expose that function via ServiceModels which can be called from the Automate Script.
I think @agrare and @Fryguy were discussing and alternative approach for the V2V problem.
Which doesn't involve any locking.

fdupont-redhat:
@mkanoor 
While I'm glad to see @agrare and @Fryguy scratching their heads to find a solution for V2V, I'm also concerned about the other use cases for locking, which won't have a solution. Is this also part of the discussion?

jeffwarnica:
A locking mechanism should be implemented to solve the general case.

agrare:
@fdupont-redhat we were concerned about exposing any locking primitives to automate code for general use.  @Fryguy can probably elaborate but writing distributed reentrant code is hard and users _will_ get deadlocks.
Coming back to v2v, if we do need to lock we'd prefer it were done in the core models, so add a conversion host model that we can handle this internally.

fdupont-redhat:
@agrare 
Ok to have the locking mechanism in the model. At least, V2V will be covered.

But as  @jeffwarnica says, the general case should also be solved. Otherwise, we'll see people using the advisory lock, like it has already been done in the aforementioned Gist. If you want to prevent deadlock, someone has to block advisory lock calls and provide a safe mechanism.

fdupont-redhat:
Ok, so we moved away from that for VM migration. As it was a global push back from core devs, I close this PR.

PR #355, opened 2018-07-10
Add support for pre/post-migration playbook

fdupont-redhat:
This PR adds support to call playbooks in pre and post migration phases. The implementation is generic to allow calling playbooks for other hooks. The backend provides methods to get the service template associated to pre and post hooks. The `LaunchPlaybookAsAService` method call `create_provision_request` and stores the request id in the task options hash. The `CheckPlaybookAsAService` method checks the status of the request. It gracefully fails if the playbook fails in post migration.

The `CheckPlaybookAsAService` method also updates the Ansible Tower job status in the task options hash, so that the UI can display it and also collect the job output.

Unfortunately, the test to check whether the destination VM is on was introduced inconsistency. It used the oVirt SDK gem to get the information from the provider. It was meant to be faster than wait for refresh. However, with targeted refresh, the power state is reported fast. So I changed the test to use `destination_vm.power_state == 'on'`. Without it the post migration playbook could be skipped due to race condition.

This PR adds add the following attributes to task options hash (example):

```
"pre_ansible_playbook_service_request_id" => 3,
:playbooks => {
    "pre": {
      :job_state => "finished",
      :job_status => "Ok",
      :job_id => 4
    },
    "post": {
      :job_state => "pending"
    }
}
```

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1564250

### DIFF
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
@@ -14,22 +14,25 @@ module ManageIQ
             transformation_hook = @handle.inputs['transformation_hook']
             task = @handle.root['service_template_transformation_plan_task']
             service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id")
-            service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
 
-            playbooks_status = task.get_option(:playbooks) || {}
-            playbooks_status[transformation_hook] = { :job_state => service_request.request_state }
+            if service_request_id.present?
+              service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
 
-            if service_request.request_state == 'finished'
-              @handle.log(:info, "Ansible playbook service request (id: #{service_request_id}) is finished.")
-              playbooks_status[transformation_hook][:job_status] = service_request.status
-              playbooks_status[transformation_hook][:job_id] = service_request.miq_request_tasks.first.destination.service_resource.first.id
-              if service_request.status == 'Error' && transformation_hook == 'pre'
-                raise "Ansible playbook has failed (hook=#{transformation_hook})"
+              playbooks_status = task.get_option(:playbooks) || {}
+              playbooks_status[transformation_hook] = { :job_state => service_request.request_state }
+
+              if service_request.request_state == 'finished'
+                @handle.log(:info, "Ansible playbook service request (id: #{service_request_id}) is finished.")
+                playbooks_status[transformation_hook][:job_status] = service_request.status
+                playbooks_status[transformation_hook][:job_id] = service_request.miq_request_tasks.first.destination.service_resource.first.id
+                if service_request.status == 'Error' && transformation_hook == 'pre'
+                  raise "Ansible playbook has failed (hook=#{transformation_hook})"
+                end
+              else
+                @handle.log(:info, "Playbook for #{transformation_hook} migration is not finished. Retrying.")
+                @handle.root['ae_result'] = 'retry'
+                @handle.root['ae_retry_interval'] = '15.seconds'
               end
-            else
-              @handle.log(:info, "Playbook for #{transformation_hook} migration is not finished. Retrying.")
-              @handle.root['ae_result'] = 'retry'
-              @handle.root['ae_retry_interval'] = '15.seconds'
             end
           rescue => e
             @handle.set_state_var(:ae_state_progress, 'message' => e.message)
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
@@ -1,0 +1,68 @@
+module ManageIQ
+  module Automate
+    module Transformation
+      module Ansible
+        class CheckPlaybookAsAService
+          def initialize(handle = $evm)
+            @handle = handle
+          end
+
+          def set_retry(message, interval = '1.minutes')
+            @handle.log(:info, "#{message}. Retrying.")
+            @handle.root['ae_result'] = 'retry'
+            @handle.root['ae_retry_interval'] = interval
+          end
+
+          def job_status(job)
+            { :job_id => job.id, :job_status => job.status }
+          end
+
+          def set_ansible_job(task, service_request, transformation_hook)
+            job = nil
+            service_request_task = service_request.miq_request_tasks.first
+            unless service_request_task.nil?
+              stack = service_request_task.destination.service_resources.first
+              job = stack.resource unless stack.nil?
+            end
+            playbooks_status = task.get_option(:playbooks) || {}
+
+            if playbooks_status[transformation_hook].nil?
+              if job.nil?
+                set_retry('Ansible job has not started yet.', '15.seconds')
+              else
+                playbooks_status[transformation_hook] = job_status(job)
+                task.set_option(:playbooks, playbooks_status)
+              end
+            end
+          end
+
+          def main
+            transformation_hook = @handle.inputs['transformation_hook']
+
+            task = @handle.root['service_template_transformation_plan_task']
+            service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id")
+            service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
+
+            set_ansible_job(task, service_request, transformation_hook)
+
+            if service_request.request_state == 'finished'
+              @handle.log(:info, "Ansible playbook service request (id: #{service_request_id}) is finished.")
+              if transformation_hook == 'pre' && service_request.status == 'Error'
+                raise "Ansible playbook has failed (hook=#{transformation_hook})"
+              end
+            else
+              set_retry('Service request is not finished yet.', '15.seconds')
+            end
+          rescue => e
+            @handle.set_state_var(:ae_state_progress, 'message' => e.message)
+            raise
+          end
+        end
+      end
+    end
+  end
+end
+
+if $PROGRAM_NAME == __FILE__
+  ManageIQ::Automate::Transformation::Ansible::CheckPlaybookAsAService.new.main
+end
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
@@ -13,7 +13,7 @@ module ManageIQ
           def main
             transformation_hook = @handle.inputs['transformation_hook']
             task = @handle.root['service_template_transformation_plan_task']
-            service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id")
+            service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id".to_sym)
 
             if service_request_id.present?
               service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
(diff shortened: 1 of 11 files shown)


fdupont-redhat:

fdupont-redhat:

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
```diff
@@ -0,0 +1,68 @@
...
+          def job_status(job)
+            { :job_id => job.id, :job_status => job.status }
+          end
+
+          def set_ansible_job(task, service_request, transformation_hook)
+            job = nil
+            service_request_task = service_request.miq_request_tasks.first
+            unless service_request_task.nil?
+              stack = service_request_task.destination.service_resources.first
+              job = stack.resource unless stack.nil?
```
bzwei: `stack` is already the ansible job instance. There is no `stack.resource` method.

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
```diff
@@ -0,0 +1,68 @@
...
+          end
+
+          def main
+            transformation_hook = @handle.inputs['transformation_hook']
+
+            task = @handle.root['service_template_transformation_plan_task']
+            service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id")
+            service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
+
+            set_ansible_job(task, service_request, transformation_hook)
```
bzwei: You should not drill down to examine the ansible job every time. You should rely on the service request_state. Only after the service completes you can fetch the ansible job status just one time.

fdupont-redhat:
@bzwei 
Thanks for the advice. Indeed it can be lighter and the state and status can be those of the request. The job is only useful when the request is finished.

mturley:
Part of the front-end support (the Overview page parts) for this will be merged with https://github.com/ManageIQ/miq_v2v_ui_plugin/pull/467 within the next day or two. The rest is already merge-ready I believe, as soon as this merges we should be all set end-to-end!

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
```diff
@@ -0,0 +1,51 @@
...
+
+              playbooks_status = task.get_option(:playbooks) || {}
+              playbooks_status[transformation_hook] = { :job_state => service_request.request_state }
+
+              if service_request.request_state == 'finished'
+                @handle.log(:info, "Ansible playbook service request (id: #{service_request_id}) is finished.")
+                playbooks_status[transformation_hook][:job_status] = service_request.status
+                playbooks_status[transformation_hook][:job_id] = service_request.miq_request_tasks.first.destination.service_resources.first.resource.id
+                task.set_option(:playbooks, playbooks_status)
+                if service_request.status == 'Error' && transformation_hook == 'pre'
```
mkanoor: @fdupont-redhat is it intentional to ignore the errors for post playbooks?

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
```diff
@@ -0,0 +1,51 @@
...
+            if service_request_id.present?
+              service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
+
+              playbooks_status = task.get_option(:playbooks) || {}
+              playbooks_status[transformation_hook] = { :job_state => service_request.request_state }
+
+              if service_request.request_state == 'finished'
+                @handle.log(:info, "Ansible playbook service request (id: #{service_request_id}) is finished.")
+                playbooks_status[transformation_hook][:job_status] = service_request.status
+                playbooks_status[transformation_hook][:job_id] = service_request.miq_request_tasks.first.destination.service_resources.first.resource.id
```
mkanoor: @fdupont-redhat This is a pretty long chain of commands where we could get a nil in the middle?

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb
```diff
@@ -0,0 +1,51 @@
...
+                end
+              else
+                @handle.log(:info, "Playbook for #{transformation_hook} migration is not finished. Retrying.")
+                @handle.root['ae_result'] = 'retry'
+                @handle.root['ae_retry_interval'] = '15.seconds'
+              end
+              task.set_option(:playbooks, playbooks_status)
+            end
+          rescue => e
+            @handle.set_state_var(:ae_state_progress, 'message' => e.message)
```
mkanoor: @fdupont-redhat do you want to report progress only when an error happens or should that also happen on success?
Does :ae_state_progress only mean error messages, if it does we should change it to something like :ae_state_error

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
```diff
@@ -0,0 +1,51 @@
...
+          def main
+            task = @handle.root['service_template_transformation_plan_task']
+            transformation_hook = @handle.inputs['transformation_hook']
+
+            unless transformation_hook == '_'
+              service_template = task.send("#{transformation_hook}_ansible_playbook_service_template")
+              @handle.log(:info, "Service Template Id: #{service_template.id}")
+              unless service_template.nil?
+                target_host = target_host(task, transformation_hook)
+                unless target_host.nil?
```
mkanoor: @fdupont-redhat 
Can this be a positive case with an 'if' statement instead of an unless, it makes it easier to read

```
 if service_template
.....
   target_host = target_host(task, transformation_hook)
   if target_host
       ......
   end
end
```
Also what happens if you can't find a target_host or service_template should we report on that?

bzwei: rubocop actually prefer earlier termination. In this example you can code this way
```
return if service_template.nil?
...
return if target_host.nil?
...
return unless target_host.power_state == 'on'
...

File: content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb (hunk 3 above)
Comment (fdupont-redhat) [reply content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb]: @mkanoor 
Yes, it is. We consider that the migration has gone far enough to not cancel it for a post migration playbook failure. The playbook or any other problem can be fixed afterwards, with a warning message. The UI knows that the playbook has failed, and it can show the log to the user.

File: content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb (hunk 4 above)
Comment (fdupont-redhat) [reply content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb]: @mkanoor 
Good catch. I had the same feeling, but aesthetics won over security. I'll add extra tests.

File: content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb (hunk 5 above)
Comment (fdupont-redhat) [reply content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb]: @mkanoor 
It's a default behavior to catch the error and log it in `:ae_state_progress`, so that it can be used by `WeightedUpdateStatus`. Here, we don't plan to add additional information because the `description` argument set in the state machine is explicit enough.

File: content/automate/ManageIQ/Transformation/Ansible.class/__methods__/checkplaybookasaservice.rb [hunk 7]
```diff
@@ -0,0 +1,51 @@
...
+
+          def main
+            transformation_hook = @handle.inputs['transformation_hook']
+            task = @handle.root['service_template_transformation_plan_task']
+            service_request_id = task.get_option("#{transformation_hook}_ansible_playbook_service_request_id")
+
+            if service_request_id.present?
+              service_request = @handle.vmdb(:miq_request).find_by(:id => service_request_id)
+
+              playbooks_status = task.get_option(:playbooks) || {}
```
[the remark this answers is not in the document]
fdupont-redhat: @bzwei 
Code changed to use `to_sym` with the substituted string.
I know the key is very long, but it makes it really explicit.

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
```diff
@@ -0,0 +1,48 @@
...
+              target_host = @handle.vmdb(:vm).find_by(:id => task.get_option(:destination_vm_id))
+            end
+            target_host
+          end
+
+          def main
+            task = @handle.root['service_template_transformation_plan_task']
+            transformation_hook = @handle.inputs['transformation_hook']
+
+            unless transformation_hook == '_'
```
bzwei:
```
return if transformation_hook == '_'
```

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
```diff
@@ -0,0 +1,48 @@
...
+              service_template = task.send("#{transformation_hook}_ansible_playbook_service_template")
+              @handle.log(:info, "Service Template Id: #{service_template.id}")
+              return if service_template.nil?
+              target_host = target_host(task, transformation_hook)
+              return if target_host.nil?
+              return unless target_host.power_state == 'on'
```
bzwei: combine into `return if target_host.nil? || target_host.power_state != 'on'`

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
```diff
@@ -0,0 +1,48 @@
...
+          end
+
+          def main
+            task = @handle.root['service_template_transformation_plan_task']
+            transformation_hook = @handle.inputs['transformation_hook']
+
+            unless transformation_hook == '_'
+              service_template = task.send("#{transformation_hook}_ansible_playbook_service_template")
+              @handle.log(:info, "Service Template Id: #{service_template.id}")
+              return if service_template.nil?
```
bzwei: if `service_template.nil?` then the previous line `service_template.id` will blow up. A simple fix is `#{service_template.try(:id)}`

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
fdupont-redhat: Fixed

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
fdupont-redhat: Fixed

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
fdupont-redhat: I removed the `@handle.log` as it is mainly used for debug.

simaishi:

```
$ git log -1
commit fb6f12ceff23a21bbdb891431ff31b43820bc8ca
Date:   Thu Jul 19 08:06:54 2018 -0400

    Merge pull request #355 from fdupont-redhat/v2v_pre_post_migration_playbook_support
    
    Add support for pre/post-migration playbook
    (cherry picked from commit f5fdc8293ac754a37674920607d7743392d94fcd)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1608420
```

PR #357, opened 2018-07-11
Update WeightedUpdateStatus to handle task cancellation and cleanup

fdupont-redhat:
This PR implements task cancellation and cleanup. @bzwei has started implementing task cancellation on the backend: https://github.com/ManageIQ/manageiq/pull/17687/files. It adds an entry in task options hash called `cancel_requested` with value `true`. So we check if this entry exists and if yes we trigger the cleanup state machine and raise to abort the current state machine.

The method has also been changed to trigger the cleanup state machine when the migration is failed, i.e. the state exits `on_error`. The cleanup state machine is the same for failure and cancellation, as cancellation can be seen as a manual failure.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -137,7 +137,6 @@ module ManageIQ
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
                     create_cleanup_request(task)
-                    task.set_option('cleanup_request_id', cleanup_request.id) unless cleanup_request.nil?
                   elsif task.get_option('cancel_requested')
                     task.message = 'Cancelled'
                     create_cleanup_request(task)
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -137,7 +137,7 @@ module ManageIQ
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
                     create_cleanup_request(task)
-                  elsif task.get_option('cancel_requested')
+                  elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).absent?
                     task.message = 'Canceled'
                     create_cleanup_request(task)
                     raise "Task cancellation requested."
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -31,7 +31,7 @@ module ManageIQ
 
             def create_cleanup_request
               sm_uri = @handle.object['cleanup_state_machine']
-              unless sm_uri.blank?
+              if sm_uri.present?
                 options = {}
                 sm_uri_array = sm_uri.split('/')
                 options[:instance_name] = sm_uri_array.pop
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -138,7 +138,7 @@ module ManageIQ
                     task.message = 'Failed'
                     create_cleanup_request(task)
                   elsif task.get_option('cancel_requested')
-                    task.message = 'Cancelled'
+                    task.message = 'Canceled'
                     create_cleanup_request(task)
                     raise "Task cancellation requested."
                   else
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -38,7 +38,7 @@ module ManageIQ
                 options[:class_name] = sm_uri_array.pop
                 options[:namespace] = sm_uri_array.join('/')
                 options[:user_id] = @handle.root['user_id']
-                request = @handle.execute("create_automation_request", options, 'admin', true)
+                request = @handle.execute("create_automation_request", options, @handle.root['user'].userid, true)
                 task.set_option(:cleanup_request_id, request.id) unless request.nil?
               end
             end
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -39,8 +39,8 @@ module ManageIQ
                 options[:namespace] = sm_uri_array.join('/')
                 options[:user_id] = @handle.root['user_id']
                 request = @handle.execute("create_automation_request", options, 'admin', true)
+                task.set_option(:cleanup_request_id, request.id) unless request.nil?
               end
-              task.set_option(:cleanup_request_id, request.id) unless request.nil?
             end
 
             def on_entry(state_hash, _, _, state_weight, state_description)
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -29,6 +29,20 @@ module ManageIQ
               percent
             end
 
+            def create_cleanup_request
+              sm_uri = @handle.object['cleanup_state_machine']
+              unless sm_uri.blank?
+                options = {}
+                sm_uri_array = sm_uri.split('/')
+                options[:instance_name] = sm_uri_array.pop
+                options[:class_name] = sm_uri_array.pop
+                options[:namespace] = sm_uri_array.join('/')
+                options[:user_id] = @handle.root['user_id']
+                request = @handle.execute("create_automation_request", options, 'admin', true).id
+              end
+              request
+            end
+
             def on_entry(state_hash, _, _, state_weight, state_description)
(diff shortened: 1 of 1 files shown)


fdupont-redhat:

fdupont-redhat:

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -122,6 +136,11 @@ def main
                   # We set the task message.
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
+                    task.set_option('cleanup_request_id', create_cleanup_request.id)
```
bzwei: The `create_cleanup_request` method already returns an `id`.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -122,6 +136,11 @@ def main
...
+                  elsif task.get_option('cancel_requested')
+                    task.message = 'Cancelled'
+                    task.set_option('cleanup_request_id', create_cleanup_request.id)
+                    raise "Task cancellation requested."
```
bzwei: What is the result of raising an error? Will it eventually end with `Failed` and trigger another `create_cleanup_request`? 

Do we need to clear option `cancel_requested`? Will you check this flag multiple times and each time call `create_cleanup_request`?

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
gmcculloug: If `create_cleanup_request` does not go through the `if sm_uri.present?` block then the return will be `nil`.  Either way this method will currently error out trying to access `create_cleanup_request.id`.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -29,6 +29,20 @@ def reconcile_states_percent(path, states)
...
+              if sm_uri.present? && task.get_option(:cleanup_request_id).absent?
+                options = {}
+                sm_uri_array = sm_uri.split('/')
+                options[:instance_name] = sm_uri_array.pop
+                options[:class_name] = sm_uri_array.pop
+                options[:namespace] = sm_uri_array.join('/')
+                options[:user_id] = @handle.root['user_id']
+                request = @handle.execute("create_automation_request", options, 'admin', true)
+              end
+              task.set_option(:cleanup_request_id, request.id) unless request.nil?
```
bzwei: move this line inside the `if ... end` block

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -29,6 +29,20 @@ def reconcile_states_percent(path, states)
...
+            def create_cleanup_request(task)
+              sm_uri = @handle.object['cleanup_state_machine']
+              if sm_uri.present? && task.get_option(:cleanup_request_id).absent?
+                options = {}
+                sm_uri_array = sm_uri.split('/')
+                options[:instance_name] = sm_uri_array.pop
+                options[:class_name] = sm_uri_array.pop
+                options[:namespace] = sm_uri_array.join('/')
+                options[:user_id] = @handle.root['user_id']
+                request = @handle.execute("create_automation_request", options, 'admin', true)
```
bzwei: can we get current user instead of default to 'admin'?

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: Fixed.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
mkanoor: @fdupont-redhat Does the user exist in the DB?
https://github.com/ManageIQ/manageiq-automation_engine/blob/bbcf066a1ebb87fc7021baff6189a32db0054765/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb#L109
Its finding the user based on the User#userid field not the User.id field.
If we have a current user in the automation we should be able to use it. $evm.root['user'].userid

fdupont-redhat: Yes, the user exists. I have created a user named `bob` with id 2. I have tried the following in rails console:

```ruby
$evm.execute("create_automation_request", { instance_name: "Test", class_name: "Raise", namespace: "StateMachines", user_id: 2 }, 'bob', true)
```

And the error is the following:

```ruby
RuntimeError: not authorized
	from /home/miq/manageiq/app/models/miq_approval.rb:66:in `user_validate'
	from /home/miq/manageiq/app/models/miq_approval.rb:14:in `approve'
	from /home/miq/manageiq/app/models/miq_request.rb:307:in `approve'
	from /home/miq/manageiq/app/models/mixins/inter_region_api_method_relay.rb:30:in `block (2 levels) in api_relay_method'
	from /home/miq/manageiq/app/models/miq_request.rb:521:in `post_create'
	from /home/miq/manageiq/app/models/miq_request.rb:494:in `create_request'
	from /home/miq/manageiq/app/models/mixins/inter_region_api_method_relay.rb:48:in `block (2 levels) in api_relay_class_method'
	from /home/miq/.gem/ruby/bundler/gems/manageiq-automation_engine-bbcf066a1ebb/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_methods.rb:110:in `create_automation_request'
	from /home/miq/.gem/ruby/bundler/gems/manageiq-automation_engine-bbcf066a1ebb/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service.rb:167:in `execute'
	from (irb):6
	from /home/miq/.gem/ruby/gems/railties-5.0.7/lib/rails/commands/console.rb:65:in `start'
	from /home/miq/.gem/ruby/gems/railties-5.0.7/lib/rails/commands/console_helper.rb:9:in `start'
	from /home/miq/.gem/ruby/gems/railties-5.0.7/lib/rails/commands/commands_tasks.rb:78:in `console'
	from /home/miq/.gem/ruby/gems/railties-5.0.7/lib/rails/commands/commands_tasks.rb:49:in `run_command!'
	from /home/miq/.gem/ruby/gems/railties-5.0.7/lib/rails/commands.rb:18:in `<top (required)>'
	from bin/rails:4:in `require'
	from bin/rails:4:in `<main>'
```

gmcculloug:
User needs to have the right permission: https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_approval.rb#L44

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: After @gmcculloug comment, I did some tests around approval and it appears that the role the user belongs to needs the Services > Requests > Operate > Approve and Deny privilege, whereas I would expect it needs Automation > Automate > Requests > Operate > Approve and Deny, as it is an automation request and not a service provision request. Because, it means we have to document why the AppOwner role requires Service > Requests > Operate > Approve and Deny privilege.

Anyway, I do agree that it should be the current user that is the requester, instead of admin. But we have to document RBAC for 5.9.4, because RBAC for Transformation is not available yet.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -122,6 +136,11 @@ def main
                   # We set the task message.
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
+                    create_cleanup_request(task)
+                  elsif task.get_option('cancel_requested')
```
bzwei: I haven't done an actual test. Will the error raised multiple times? do you want to make it
```
elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).absent?
```

again here we have a mix use of symbol and string. `'cancel_requested'` should be `:cancel_requested`. When I created the PR I did not pay attention to what existing keys are. We can leave it as is since this is a temporary solution.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -122,6 +136,11 @@ def main
...
+                    task.message = 'Cancelled'
```
bzwei: Maybe `Canceled`? I am used to writing `Cancelled` too, but it seems in the US `Canceled` is more common.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: Yes, it will call `create_cleanup_request` multiple times, but it will do nothing and return nil.It doesn't seem to add overload. I'd say "as you wish" ;)

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: Yes. You're right about it. I've grep'd the code and everywhere it's `canceled`.

simaishi:

```
$ git log -1
commit aca058f46b5027f31fc8c6685c68555424015911
Date:   Fri Jul 20 07:50:43 2018 -0400

    Merge pull request #357 from fdupont-redhat/v2v_task_cancellation_support
    
    Update WeightedUpdateStatus to handle task cancellation and cleanup
    (cherry picked from commit 68c1a431898448b0069654aba3816c2a94b06066)
    
```

PR #359, opened 2018-07-12
Refactor to add support for SSH transport method

fdupont-redhat:
The original goal of this PR is to add support for SSH transport method when migrating a VM from VMware to RHV. To do that, the main change to implement is providing a different virt-v2v-wrapper options hash. Instead of adding an if statement, I created a facade method that calls the right method based on `transformation_type` (e.g. *vmware2rhevm*) and `transformation_method` (e.g. *vddk*). This way I was able to move the specific implementations to `/Transformation/TransformationHosts/Common/Utils`.

Then, the only specific code in `/Transformation/TransformationHosts/ovirt_host/VMTransform_vmware2rhevm_vddk` was the execution of remote command to triger `virt-v2v-wrapper`. As we will have more than one transformation host type, it made sense to make this call generic using another facade method in `/Transformation/TransformationHosts/Common/Utils`. This time, the actual implementation exists in the specific transformation host type class, so I use `constantize` to dynamically calling the right method in the facade method.

With these two changes, `/Transformation/TransformationHosts/ovirt_host/VMTransform_vmware2rhevm_vddk` becomes generic and it makes sense to move it to `/Transformation/TransformationHosts/Common` and to rename it to `VMTransform`.

The same logic happened is valid for `/Transformation/TransformationHosts/ovirt_host/CheckVMTransform_vmware2rhevm_vddk`, which is also moved to `/Transformation/TransformationHosts/Common` and to renamed to `CheckVMTransform`.

Then, to call the moved and renamed methods, the state machine had to be modified. I piggy backed a more complex change to the state machines by splitting it in 4:

- 1 master state machine
- 3 nested state machines for each phase of the migration: PreTransform, Transform and PostTransform

This change also modifies the weight of all the states to be consistent. A beneficial side effect is that the size of the state machines is reduced to 20 states, including empty states dedicated to customization.

Another change concerns conversion host acquisition, because transport method known from the tag the selected host has. So, we don't filter the hosts based on the transport method anymore, but we return the transport method of the conversion host. It is expected that they have a consistent method across a provider.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1564251


fdupont-redhat:

fdupont-redhat:

bthurber:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1564251

fdupont-redhat:

bzwei:
According to the description the refactoring is on the right direction 👍

bthurber:
@fdupont-redhat looks like a rebase is needed before merge will complete.

fdupont-redhat:
@miq-bot remove-label wip

fdupont-redhat:
@miq-bot remove-label unmergeable

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/vmtransform.rb
```diff
@@ -0,0 +1,70 @@
...
+                # Record the wrapper files path
+                @handle.log(:info, "Command stdout: #{result[:stdout]}") if @debug
+                task.set_option(:virtv2v_wrapper, JSON.parse(result[:stdout]))
+
+                # Record the status in the task object
+                task.set_option(:virtv2v_started_on, start_timestamp)
+                task.set_option(:virtv2v_status, 'active')
+              end
+
+              if task.get_option(:virtv2v_started_on).nil?
```
mkanoor: @fdupont-redhat 
Can you convert the retry logic into a separate function (e.g. set_retry) and call  set_retry after line 31

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/vmtransform.rb
```diff
@@ -0,0 +1,70 @@
...
+                @handle.root['ae_result'] = 'retry'
+                @handle.root['ae_retry_server_affinity'] = true
```
mkanoor: @fdupont-redhat Do you want to retry back from this server even if other servers are free to do the work?

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -31,20 +33,85 @@ def self.transformation_hosts(ems, method, factory_config)
...
+            def self.get_runners_count_by_ems(ems, factory_config)
+              transformation_hosts(ems, factory_config).inject(0) { |sum, thost| sum + thost[:runners][:current] }
             end
 
-            def self.get_transformation_host(ems, method, factory_config)
-              ems_max_runners = ems.custom_get('Max Transformation Runners') || factory_config['ems_max_runners'] || 1
-              ems_cur_runners = get_runners_count_by_ems(ems, method, factory_config)
-              transformation_host = ems_cur_runners < ems_max_runners ? eligible_transformation_hosts(ems, method, factory_config).first[:host] : nil
-              transformation_host
+            def self.get_transformation_host(task, factory_config, handle = $evm)
```
bzwei: I would prefer to revert `task` to `ems`. As a utility method it is better to accept more direct objects. Same for other methods in this class where `task` is accepted.

content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
```diff
@@ -31,6 +24,8 @@ def main
               @handle.log(:info, "Transformation Host: #{transformation_host.name}.")
               task.set_option(:transformation_host_id, transformation_host.id)
               task.set_option(:transformation_host_name, transformation_host.name)
+              task.set_option(:transformation_host_type, transformation_host_type)
```
bzwei: do you want to group all `transformation_host` related options under key `:transformation_host`?

content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
```diff
@@ -53,7 +53,9 @@ def main
             raise "No destination cluster for '#{source_vm.name}'. Exiting." if destination_cluster.nil?
 
             source_ems = source_vm.ext_management_system
+            task.set_option(:source_ems_id, source_ems.id)
```
bzwei: Do you really need to store `:source/destination_ems_id`? They can be easily retrieved though one or few method calls. The `options` field is getter fatter and fatter. We should store here information that we cannot get from other places, for example the progress.

bzwei: You can create utility methods for procedures like the example you gave. Better we can move this code to backend model and its corresponding service model. 
We can merge the PR as is if it is urgently needed, but I prefer a follow-up refactoring to move some redundant information out of the options field. They should be replaced by a dedicated method.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
bzwei: ok

content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
fdupont-redhat: Probably a topic for `options` hash refactoring, isn't it ?

content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
fdupont-redhat: I agree that taking time to rethink the options hash and moving thing in the ServiceTemplateTransformationPlanTask model would be good. We'll do it for gaprindashvili-5 ?

gmcculloug:
@bzwei @mkanoor Looks like we are getting close on this one, please mark the PR as approved when you think it is ready to merge.  Thanks.

simaishi:

```
$ git log -1
commit ba2c204a76cf249b41aa63d814d0e674e28c0140
Date:   Wed Jul 25 14:45:22 2018 -0400

    Merge pull request #359 from fdupont-redhat/v2v_refactor_vmtransform_method_for_ssh_support
    
    Refactor to add support for SSH transport method
    (cherry picked from commit a7035b74981613ef2b242d646c38568c24e09371)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1610055
```

Issue #350, opened 2018-07-04
Move 'dump_root' logic into common method location and include as embedded method


gmcculloug:
The basic "dump_root" logic from the [InspectMe method](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/__methods__/inspectme.rb#L7-L12
) is used in several methods:

```
ManageIQ/Cloud/Orchestration/Retirement/StateMachines/Methods.class/__methods__/start_retirement.rb
ManageIQ/Control/Email.class/__methods__/parse_alerts.rb
ManageIQ/Infrastructure/VM/Retirement/StateMachines/Methods.class/__methods__/start_retirement.rb
ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/check_provisioned.rb
ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/groupsequencecheck.rb
ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/provision.rb
```

The logic should be moved into a new method which can then be included into the other methods using the Embedded method feature.  The InspectMe method should also include this logic through the embedded method as well.

Potentially all the methods in InspectMe could be exposed in a utility class for reuse and the InspectMe method today would just become simple calls to those embedded methods.

cc @tinaafitz @mkanoor

gmcculloug:
Issue resolved with https://github.com/ManageIQ/manageiq-content/pull/386

PR #362, opened 2018-07-16
Created DumpRoot instance and method in System/Request class.

billfitzgerald0120:
This is a new method which can then be included into the other methods using the Embedded method feature.

https://github.com/ManageIQ/manageiq-content/issues/350

@miq-bot assign @gmcculloug


content/automate/ManageIQ/System/Request.class/__methods__/dump_root.rb
```diff
@@ -0,0 +1,12 @@
...
+#
+
+def dump_root
+  $evm.log("info", "Root:<$evm.root> Attributes - Begin")
+  $evm.root.attributes.sort.each { |k, v| $evm.log("info", "  Attribute - #{k}: #{v}") }
+  $evm.log("info", "Root:<$evm.root> Attributes - End")
+  $evm.log("info", "")
+end
+
+dump_root
```
gmcculloug: Since the goal is to use this as an embedded method we do not want the method to be called from here.  The automate method that embeds it would inject the call where needed.

Also need to consider moving it to another namespace in Automate where other common embedded methods would live.

cc @tinaafitz @mkanoor

tinaafitz: @billfitzgerald0120 Can we add a new class for these types of utilities in /System/CommonMethods/Utils?

tinaafitz: @billfitzgerald0120 The method should have the module information.
module ManageIQ
  module Automate
    module System
      module CommonMethods
        module Utils
          ........

content/automate/ManageIQ/System/Request.class/__methods__/dump_root.rb
```diff
@@ -0,0 +1,12 @@
+#
+# Description: Log all attributes stored in the $evm.root hash.
+#
+
+def dump_root
```
mkanoor: @billfitzgerald0120 to aid in testing we need to be able to inject dependencies. we won't have a $evm when we test 
so this might be a better way of accomplishing the same thing
```
def dump_root(handle = $evm)
    handle.log("info", "Root:<handle.root> Attributes - Begin")
    handle.root.attributes.sort.each { |k, v| handle.log("info", "  Attribute - #{k}: #{v}") }
    handle.log("info", "Root:<handle.root> Attributes - End")
    handle.log("info", "")
 end
```

mkanoor:
@billfitzgerald0120 can you add a spec for this.
You can look at this https://gist.github.com/jeffwarnica/bfc885ab7bd7197d626a55fa81913cc5 for some ideas on how to implement it

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/dump_root.rb
```diff
@@ -0,0 +1,10 @@
+#
+# Description: Log all attributes stored in the $evm.root hash.
+#
+
+def dump_root(handle = $evm)
```
mkanoor: @billfitzgerald0120 this needs the modules and also the name of a class preferably Utils.
And the dump_root should be a class method

```
def self.dump_root(handle = $evm)
....
end
```

gmcculloug:
@billfitzgerald0120 The branch this is built on is `dump_root` and the PR title use to reference that, but now this is all logging methods.  I would have expected this to be a new PR to add more common functionality, not remove the dump_root logic.  I'm confused.  😕

billfitzgerald0120:
@gmcculloug @mkanoor @tinaafitz  Madhu and I changed dump_root to log_object yesterday.  I will change this to WIP.  We need to discuss with Tina and Madhu.

gmcculloug:
I totally misread the code change, I see now what you are doing.  Sorry about that.  😊

billfitzgerald0120:
@tinaafitz Please Review

PR #363, opened 2018-07-18
Adjust Automate Instance names to new event type prefix

miha-plesko:
This is a followup PR for https://github.com/ManageIQ/manageiq-providers-nuage/pull/84 where Nuage event types were added `nuage_` prefix so they don't match Automate Instance definitions anymore, hence targeted refresh doesn't work.

@miq-bot assign @gmcculloug

/cc @agrare @Ladas


gmcculloug:
@agrare @Ladas Looks good to me.  Just looking for one of you to give the approval.

PR #365, opened 2018-07-19
Actually rename AE Instances, not just fixture filename

miha-plesko:
With previouse commit we accidently only updated filenames leaving actual AE Instance names without the `nuage_` prefix. Fixed.

Followup for: https://github.com/ManageIQ/manageiq-content/pull/363

@miq-bot assign @gmcculloug 

/cc @agrare @Ladas


miha-plesko:
Guys sorry, I should have double checked in first place. Today I tested on fresh database and noticed `nuage_` prefix is still missing.

PR #367, opened 2018-07-23
Trigger targeted refresh for new Nuage entities

miha-plesko:
With this commit we provide default AE Instances that are responsible of triggering event-based targeted refresh on recently supported Nuage entities. Namely, we support:

- sharednetworkresource (aka CloudNetwork)
- l2domain (aka CloudSubnet::L2)
- floatingip (aka FloatingIp)
- vport (aka NetworkPort)

@miq-bot assign @gmcculloug 

/cc @Ladas @agrare


miha-plesko:
With this commit we make sure that all the `when` statements from [here](https://github.com/ManageIQ/manageiq-providers-nuage/blob/master/app/models/manageiq/providers/nuage/network_manager/event_target_parser.rb#L26-L39) are possible to happen.

PR #377, opened 2018-07-27
Fix weightedupdatestatus for cancellation

fdupont-redhat:
Apparently and contrary to my tests, the attributes in a state machine are available in $evm.root and not in $evm.object. To work correctly, I grab the cleanup state machine path from $evm.root.

Also, I realized that `nil.absent?` generates an exception. Fixing that.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -30,8 +30,8 @@ module ManageIQ
             end
 
             def create_cleanup_request(task)
-              sm_uri = @handle.object['cleanup_state_machine']
-              if sm_uri.present? && task.get_option(:cleanup_request_id).absent?
+              sm_uri = @handle.root['cleanup_state_machine']
+              if sm_uri.present? && task.get_option(:cleanup_request_id).nil?
                 options = {}
                 sm_uri_array = sm_uri.split('/')
                 options[:instance_name] = sm_uri_array.pop
@@ -137,7 +137,7 @@ module ManageIQ
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
                     create_cleanup_request(task)
-                  elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).absent?
+                  elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).nil?
                     task.message = 'Canceled'
                     create_cleanup_request(task)
                     raise "Task cancellation requested."
--- a/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
@@ -31,7 +31,7 @@ module ManageIQ
 
             def create_cleanup_request(task)
               sm_uri = @handle.root['cleanup_state_machine']
-              if sm_uri.present? && task.get_option(:cleanup_request_id).nil?
+              if sm_uri.present? && task.get_option(:cleanup_request_id).blank?
                 options = {}
                 sm_uri_array = sm_uri.split('/')
                 options[:instance_name] = sm_uri_array.pop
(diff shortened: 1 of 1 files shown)


fdupont-redhat:
@miq-bot add-label bug
@miq-bot add-label blocker

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -30,8 +30,8 @@ def reconcile_states_percent(path, states)
             end
 
             def create_cleanup_request(task)
-              sm_uri = @handle.object['cleanup_state_machine']
-              if sm_uri.present? && task.get_option(:cleanup_request_id).absent?
+              sm_uri = @handle.root['cleanup_state_machine']
+              if sm_uri.present? && task.get_option(:cleanup_request_id).nil?
```
gmcculloug: I would suggest using `.blank?`

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -137,7 +137,7 @@ def main
                   if @handle.root['ae_state_step'] == 'on_error'
                     task.message = 'Failed'
                     create_cleanup_request(task)
-                  elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).absent?
+                  elsif task.get_option('cancel_requested') && task.get_option(:cleanup_request_id).nil?
```
gmcculloug: Same

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: Followed your suggestion.

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
fdupont-redhat: Same ;)

content/automate/ManageIQ/System/CommonMethods/MiqAe.class/__methods__/weightedupdatestatus.rb
```diff
@@ -30,8 +30,8 @@ def reconcile_states_percent(path, states)
             end
 
             def create_cleanup_request(task)
-              sm_uri = @handle.object['cleanup_state_machine']
-              if sm_uri.present? && task.get_option(:cleanup_request_id).absent?
+              sm_uri = @handle.root['cleanup_state_machine']
```
gmcculloug: @fdupont-redhat Can you describe this change in more detail?  How is 'cleanup_state_machine' being resolved into the automate workspace?

cc @mkanoor

fdupont-redhat: `cleanup_state_machine` is an attribute of the main state machine `/Transformation/StateMachines/VMTransformation/Transformation` [1]. At first, I thought it would be an attribute of `$evm.object`, but dumping both `$evm.object` and `$evm.root` showed it was only available as an attribute of `$evm.root`. I also saw that it persists in the embedded state machines, which is good.

[1] https://github.com/fdupont-redhat/manageiq-content/blob/865897e21452a423c6769d869af3eb256ef4774d/content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/__class__.yaml#L14-L33

mkanoor: @fdupont-redhat 
The cleanup_state_machine is in the first state machine that starts which might be the root object. Then you go into nested state machines and each of the state machines ends up calling the same weightedupdatestatus method, if you are in the lower level state machines that cleanup_state_machine will not percolate down into the individual state machines, they would have their own $evm.object which would be the child state machine. You would have to check the $evm.root for this attribute or check if the weighted_status is being called from the correct state machine.

fdupont-redhat: Indeed cleanup_state_machine is available at the lowest level state machine, which creates the cleanup request. The parent state machines will also fail, but won't create additional requests because we check if a request already exists.

simaishi:

```
$ git log -1
commit 3a98ce68ea95d6d5e4a6468c9fa1fbd04babbcf1
Date:   Mon Jul 30 16:50:31 2018 -0400

    Merge pull request #377 from fdupont-redhat/v2v_fix_weightedupdate_status_for_cancellation
    
    Fix weightedupdatestatus for cancellation
    (cherry picked from commit 100d045042d67455fe5e89893863b3c1f199beec)
    
```

PR #378, opened 2018-07-28
Add default cleanup state machine for VM transformation

fdupont-redhat:
This PR provides the default state machine for VM transformation cleanup. The state machine is triggered from an automate request, as designed in https://github.com/ManageIQ/manageiq-content/pull/357. The request creation passes a variable in `$evm.root`: `service_template_transformation_plan_task_id`.

The process is as follow:

1. Assess the transformation cleanup. It basically sets the required state vars.
2. Kill virt-v2v. The PID of the process and the transformation host information is available in the task. It simply needs a remote command execution to kill the process. `virt-v2v-wrapper` handles cleaning up the destination provider when the virt-v2v command fails, so nothing to do on ManageIQ side.
3. Power on the VM if it was powered on when the migration started.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1599997

PR #382, opened 2018-07-31
Fix default cleanup state machine

fdupont-redhat:
While looking to refactor `CheckPoweredOn` state, I realized that the cleanup state machine was targeting the destination VM for power on, while it should target source VM. The logic is similar to what is implemented in `poweron.rb` in #378. This PR implements it in `checkpoweredon.rb`.

Doing so makes the `checkpoweredon.rb` method generic, so it's moved to `/Transformation/Infrastructure/VM/Common` class. This requires changing two state machines: `PostTransform` and `TransformationCleanup`.

Changing `TransformationCleanup` state machine removes the need for `destination_ems_type` state var. The method `assesstransformationcleanup.rb` only purpose was to set this state var, so it can be removed.

When running tests, I also realized that `service_template_transformation_plan_task_id` attribute was missing in `$evm.root`. Somehow, the change is `weightedupdatestatus.rb` to add it to the cleanup request was not pushed in #378, so I'm fixing it here.


fdupont-redhat:
@miq-bot add-label bug
@miq-bot add-label blocker

content/automate/ManageIQ/Transformation/Infrastructure/VM/Common.class/__methods__/checkpoweredon.rb
```diff
@@ -0,0 +1,41 @@
...
+              end
+
+              def main
+                if @handle.root['service_template_transformation_plan_task'].blank?
+                  task = @handle.vmdb(:service_template_transformation_plan_task).find_by(:id => @handle.root['service_template_transformation_plan_task_id'])
+                  vm = task.source if task.present?
+                else
+                  task = @handle.root['service_template_transformation_plan_task']
+                  vm = @handle.vmdb(:vm).find_by(:id => task.get_option(:destination_vm_id)) if task.present?
+                end
```
gmcculloug: We should break this down into methods which would make the logic easier to read.

Move this first if-else block into a method that returns both `task` and `vm`:
```ruby
def task_and_vm
  if @handle.root['service_template_transformation_plan_task'].blank?
    task = @handle.vmdb(:service_template_transformation_plan_task).find_by(:id => @handle.root['service_template_transformation_plan_task_id'])
    vm = task.source if task.present?
  else
    task = @handle.root['service_template_transformation_plan_task']
    vm = @handle.vmdb(:vm).find_by(:id => task.get_option(:destination_vm_id)) if task.present?
  end

  return task, vm
end
```

Then the first part of main becomes:
```ruby
task, vm = task_and_vm
```

If you think this is too much change it could be in a followup refactoring, but we need to move away for having large single method scripts in general.

content/automate/ManageIQ/Transformation/Infrastructure/VM/Common.class/__methods__/checkpoweredon.rb
```diff
@@ -0,0 +1,41 @@
...
+                return if vm.blank?
+                @handle.log(:info, "Target VM: #{vm.name} [#{vm.vendor}]")
+                return if task.get_option(:source_vm_power_state) != 'on'
+                if vm.power_state != 'on'
```
gmcculloug: For readability add spaces between lines when the context changes:

I would think something like this:
```ruby
def main
  task, vm = task_and_vm
  return if vm.blank?

  @handle.log(:info, "Target VM: #{vm.name} [#{vm.vendor}]")

  return if task.get_option(:source_vm_power_state) != 'on'

  if vm.power_state != 'on'
    @handle.root["ae_result"] = "retry"
    @handle.root["ae_retry_interval"] = "15.seconds"
  end
rescue => e
  @handle.set_state_var(:ae_state_progress, 'message' => e.message)
  raise
end
```

content/automate/ManageIQ/Transformation/Infrastructure/VM/Common.class/__methods__/checkpoweredon.rb
fdupont-redhat: The code is identical in `checkpoweredon.rb` and `poweron.rb`. I'd propose moving `task_and_vm` to a embedded utility method. It's a bigger change that may require another PR. I'd go for a followup refactoring.

simaishi:

```
$ git log -1
commit 3d43a639999bb0194e34c8a80a7d6f360dfaa32e
Date:   Tue Jul 31 20:48:51 2018 -0400

    Merge pull request #382 from fdupont-redhat/v2v_fix_cleanup_state_machine_poweron
    
    Fix default cleanup state machine
    (cherry picked from commit 3e6a3f656cac57ad1b9ec2c1a6d1e8cdde44e6f4)
    
```

PR #387, opened 2018-08-02
Don't sleep 3.minutes

bdunne:
wat_owl


jrafanie:
WOW.  😮  OOPS

That sleep Drew my attention.

PR #388, opened 2018-08-02
Fix quota requested cloud_volume calculations.

tinaafitz:
Add spec test for google cloud volumes and check for cloud_volumes unconditionally.

The quota requested method was checking for cloud_volumes only on a reconfigure request. We should be checking for cloud_volumes all cloud provisioning requests.


tinaafitz:
@miq-bot assign @mkanoor 
@miq-bot add_label bug

tinaafitz:
@billfitzgerald0120 Please review.

PR #390, opened 2018-08-02
Remove the `if` conditionals to improve test coverage

bdunne:
Prevent the module from running when required by hooking the class definition and stubbing the call to `.new` during require.  Doing this ensures that the `.new` calls a real class that already exists.

Unfortunately, the class name is occasionally misspelled and this code path was previously not covered.  This adds test coverage for this path.


mkanoor:
@bdunne 
This skews the coverage and now only covers the calling of block around the main. All other data is lost

bdunne:
Okay, updated the PR with the new approach to increasing test coverage.  Let me know what you think @mkanoor 
There are a few classes outside of the "ManageIQ::Automate" namespace that are not covered by this change, but I think those should be addressed by changing the namespacing in a separate PR.
cc @Fryguy

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
```diff
@@ -63,6 +63,4 @@ def fill_dialog_field(list)
   end
 end
 
-if __FILE__ == $PROGRAM_NAME
-  ManageIQ::Automate::AutomationManagement::AnsibleTower::Operations::AvailableCredentials.new.main
-end
+ManageIQ::Automate::AutomationManagement::AnsibleTower::Operations::AvailableCredentials.new.main
```
jrafanie: Is that the intention?  To instantiate this class unconditionally?

bdunne: Yes.  In test, https://github.com/ManageIQ/manageiq-content/pull/390/files#diff-93830fa29d616f7c87903d08b5b1b29aR66 will catch this on load and raise if the constant doesn't exist.  I don't think there is any other use case for this conditional.

spec/support/shared/examples/drew_test_spec.rb
```diff
@@ -1,14 +0,0 @@
-RSpec.shared_examples "automate_engine_call" do |filename|
-  it "drew test" do
```
jrafanie: `it "drew my attention" do`

jrafanie: lol

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/Methods.class/__methods__/available_credentials.rb
jrafanie: What about non-test code?   Does this file get required ever?

bdunne: @mkanoor can you confirm that this is never `require`d or `load`ed?

mkanoor: @bdunne That if conditional is to only allow the method to be called when its executed from the Automate engine. If we remove the condition it breaks that logic.

bdunne: @mkanoor The two cases I know of are:
- Tests: we don't want to run the code
- Execution: we want the code to run in the automation engine

Are there any other cases where these files are required or loaded and we don't want it to run?

mkanoor: @bdunne 
During testing we only want to run the code when we have a test setup done, so that the invocation doesn't fail. That would give us the 100% coverage. Today we test everything for these new methods except for the if block code.

Not all newer classes have the If Block.

spec/spec_helper.rb
```diff
@@ -45,3 +45,45 @@ def domain_file
...
+      hooked << klass
+      klass.define_singleton_method(:new) do |*_args|
+        AutomateClassDefinitionHook::TestLoadStub.new
+      end
+    end
+  end
+
+  def self.unhook
+    hooked.delete_if do |klass|
+      klass.singleton_class.redefine_method(:new) { |*args| super(*args) }
```
Fryguy: Prefer `remove_method` instead of `redefine_method` to call super.  This way it goes back to the original implementation.

PR #396, opened 2018-08-09
ManageIQ and the default automate domain don't require the savon gem

bdunne:
https://bugzilla.redhat.com/show_bug.cgi?id=1573658

### DIFF
--- a/manageiq-content.gemspec
+++ b/manageiq-content.gemspec
@@ -15,8 +15,6 @@ Gem::Specification.new do |s|
 
   s.files = Dir["{content,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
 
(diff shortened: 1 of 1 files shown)


bdunne:
@gmcculloug Yes, I know, but there are RFE BZ's to update the version of the dependency.  So, rather than updating the version and breaking customers in unexpected ways, I'd rather remove the dependency entirely (since our code does not depend on it).

Since it is not our dependency it creates a number of questions:
- What version do we upgrade to?
  - Do we sample a group of people and find the closest version that they can agree on?
  - How do we communicate the change?
- Do we never upgrade?
- If there is a security issue, are we stuck trying to backport fixes to an unmaintained version?

PR #398, opened 2018-08-13
Add CustomButtonEvent to automate explorer

tumido:
Adding `CustomButtonEvent` to the automate explorer.

Related: https://github.com/ManageIQ/manageiq/pull/17764

cc @lfu Can you help me please? Is this all I need to do? Or should I create `System/Event/CustomButtonEvent.class` folder instead? If so, how do I generate the `__class__.yaml` file?

@miq-bot add_label enhancement, events

### DIFF
--- a/content/automate/ManageIQ/System/Event/CustomButtonEvent/UI.class/button.trigger.start.yaml
+++ b/content/automate/ManageIQ/System/Event/CustomButtonEvent/UI.class/button.trigger.start.yaml
@@ -1,0 +1,10 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: button.trigger.start
+    inherits: 
+    description: 
+  fields: []
--- a/content/automate/ManageIQ/System/Event/CustomButtonEvent/__namespace__.yaml
+++ b/content/automate/ManageIQ/System/Event/CustomButtonEvent/__namespace__.yaml
@@ -1,0 +1,10 @@
+---
+object_type: namespace
+version: 1.0
+object:
+  attributes:
+    name: CustomButtonEvent
+    description: 
+    display_name: 
+    priority: 
+    enabled: 
--- a/content/automate/ManageIQ/System/Event/CustomButtonEvent/UI.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/CustomButtonEvent/UI.class/__class__.yaml
@@ -1,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: 
+    name: UI
+    type: 
+    inherits: 
+    visibility: 
+    owner: 
+  schema:
+  - field:
+      aetype: assertion
+      name: guard
+      display_name: 
+      datatype: string
+      priority: 1
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: attribute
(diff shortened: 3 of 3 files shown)


lfu:
@tumido Please follow [this guide](http://talk.manageiq.org/t/creating-a-pull-request-for-enhancements-to-manageiq-automate-domain/1041) when working on automate modeling.

We need to add a new namespace ```CustomButtonEvent``` and new classes for all ```source``` values. You can copy class schema from existing ones then rename it. New instances can be added for each class based on ```event_type``` value.

tumido:
@lfu I'm finally starting to understand the workflow! :wink: And now I see why it doesn't make much sense to have the `event_type` named `custom_event_button` as you suggested in the core PR.

What would you say if we rename it to `button_triggered` or something like that? So it would look like:

- `System/Event` has a new namespace `CustomButtonEvent`
- `CustomButtonEvent` has a new class `UI`
- in this class there is a new instance `button_triggered` (the `event_type` would be adjusted in core of course)

It's much better readable and easier to follow to me. What do you think?

lfu:
@tumido The new structure layout looks good. 👍 
[Peter's book](https://pemcg.gitbooks.io/mastering-automation-in-cloudforms-4-2-and-manage/content/introduction_to_the_automate_datastore/chapter.html) can help you get familiar with automate. Just go through the first couple of chapters, specially about coping a class.

The new class is not correct. It should be similar to [this class](https://github.com/ManageIQ/manageiq-content/blob/9e53dd12d0b642452d732513fc454903b92cef4b/content/automate/ManageIQ/System/Event/EmsEvent/VC.class/__class__.yaml).  We can have a bluejeans session if you need help.

content/automate/ManageIQ/System/Event/CustomButtonEvent/__namespace__.yaml
```diff
@@ -0,0 +1,10 @@
+---
+object_type: namespace
+version: 1.0
+object:
+  attributes:
+    name: CustomButtonEvent
+    description: ''
```
lfu: Please leave the description empty if you don't have one.

tumido:
@lfu, can you please review the scheme now? Also I'm not sure if I should have kept the `message` field the same as for the example you've linked before.

Thanks for the link to the book, btw. :+1:  It was helpful!

lfu:
Other than the new event_type ```button_triggered```, it looks good to me 👍

tumido:
@lfu haha, what do you mean other than the new `event_type`? What would you suggest to name it? :wink: I'd gladly change it to anything which is more likable and represents the event well.

lfu:
@tumido Are we going to take any action in automate when the event comes in? If we won't do anything based on the event in automate, then we can use the ```.missing``` instance like [this](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Event/EmsEvent/VC.class/_missing.yaml) which works as a default for any ```event_type``` that are not listed under the class.

gmcculloug:
@lfu There are no default actions required on these events at the moment.  However, since we are only emitting a single event right now I would prefer to keep the event name tied to the instance name, otherwise it is difficult for users to know what events would be expected.  The `.missing` is great when we do not control the source and there are potentially unexpected event types we need to process.

Maybe there is a better way to solve this, not sure.

content/automate/ManageIQ/System/Event/CustomButtonEvent/UI.class/button.trigger.start.yaml
```diff
@@ -0,0 +1,10 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: button_triggered
```
lfu: You forgot to rename it here.

lfu:
@tumido The last commit was manually modified. Please follow the guide. Thanks.

Issue #360, opened 2018-07-13
Implement specs for transformation methods


fdupont-redhat:
Over time, the Transformation namespace has been populated to implement new features, but no spec has been provided, increasing the overall technical debt of the project. This issue tracks the implementation of specs for this namespace, in order to reach full coverage:

- [x] [[PR#441]](https://github.com/ManageIQ/manageiq-content/pull/441) /Transformation/Common.class/__methods__/acquiretransformationhost.rb
- [x] [[PR#408]](https://github.com/ManageIQ/manageiq-content/pull/408) /Transformation/Common.class/__methods__/assesstransformation.rb
- [X] [[PR#385]](https://github.com/ManageIQ/manageiq-content/pull/385) /Transformation/Infrastructure/VM/Common.class/__methods__/restorevmattributes.rb
- [x] [[PR#391]](https://github.com/ManageIQ/manageiq-content/pull/391) [[PR#410]](https://github.com/ManageIQ/manageiq-content/pull/410) /Transformation/Infrastructure/VM/Common.class/__methods__/utils.rb
- [x] [[PR#441]](https://github.com/ManageIQ/manageiq-content/pull/441) /Transformation/Infrastructure/VM/rhevm.class/__methods__/checkpoweredon.rb
- [x] [[PR#399]](https://github.com/ManageIQ/manageiq-content/pull/399) /Transformation/Infrastructure/VM/Common.class/__methods__/checkvmininventory.rb
- [x] [[PR#421]](https://github.com/ManageIQ/manageiq-content/pull/421) /Transformation/TransformationHosts/Common.class/__methods__/utils.rb
- [x] [[PR#433]](https://github.com/ManageIQ/manageiq-content/pull/433) /Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v.rb
- [x] [[PR#421]](https://github.com/ManageIQ/manageiq-content/pull/421) /Transformation/TransformationHosts/ovirt_host.class/__methods__/utils.rb

**Previously specified methods that have been removed from the repo:**
/Transformation/Infrastructure/VM/Common.class/__methods__/poweroff.rb
/Transformation/Infrastructure/VM/Common.class/__methods__/poweron.rb
/Transformation/Infrastructure/VM/vmwarews.class/__methods__/collapsesnapshots.rb
/Transformation/Infrastructure/VM/vmwarews.class/__methods__/setmigrated.rb
[...]

PR #399, opened 2018-08-13
Refactor and add specs for Transformation - CheckVMInInventory

fdupont-redhat:
The PR also adds the specs for this method, according to https://github.com/ManageIQ/manageiq-content/issues/360.

When writing the method, it was clear that checking for virtual machine in the VMDB is not specific to the provider. This PR refactors the code to make it generic and split the code in up-to-5-lines methods. It also means moving the method and updating the Transformation state machine.

Associated BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1600870


fdupont-redhat:
@miq-bot add-label blocker

simaishi:

PR #406, opened 2018-08-21
Prepare Nuage authentication attributes (for Embedded Ansible)

miha-plesko:
Since there is no AWX credential type for Nuage, we're not able to use standard Embedded Ansible authentication approach. As a workaround we therefore set the five AE Class Attributes:

- nuage_username
- nuage_password
- nuage_enterprise
- nuage_url
- nuage_api_version

Playbooks then fetch them from AE Instance, by means of following role (available on Ansible Galaxy):

https://github.com/xlab-si/ansible-role-nuage-miq-automate
https://galaxy.ansible.com/xlab_si/nuage_miq_automate

With this commit we babysit user who will now only have to provide values (username, password, etc.) while AE attributes will already exist and will be in appropriate sequence.

image

@miq-bot assign @gmcculloug

/cc @Ladas @agrare

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/NUAGE.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/NUAGE.class/__class__.yaml
@@ -33,10 +33,110 @@ object:
       max_time: 
   - field:
       aetype: attribute
-      name: logical_event
+      name: nuage_username
       display_name: 
       datatype: string
       priority: 2
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: attribute
+      name: nuage_password
+      display_name: 
+      datatype: password
+      priority: 3
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: attribute
+      name: nuage_enterprise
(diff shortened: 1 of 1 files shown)


gmcculloug:
@miha-plesko Please review the test failure comments to get this PR green.  https://travis-ci.org/ManageIQ/manageiq-content/jobs/418651250#L1946

miha-plesko:
Thanks @gmcculloug green.

gmcculloug:
@Ladas @agrare Can you review/approve this PR as from the automate side of things this it is just a schema change.

agrare:
@miha-plesko so this task specifically gets these values right? https://github.com/xlab-si/ansible-role-nuage-miq-automate/blob/master/tasks/authentication.yaml

@gmcculloug looks okay to me, looks like this is just setting some variables in automate then retrieving their values to be used later.

miha-plesko:
@agrare yes, and it forms a hash consumable by Nuage Ansible Module, see here https://docs.ansible.com/ansible/latest/modules/nuage_vspk_module.html#parameters, the "auth" parameter.

miha-plesko:
So the idea is that playbooks that will be used via Embedded Ansible just require the role and assume `nuage_auth` hash is filled with values e.g. like this:

```yaml
---
- name: Nuage event callback
  roles:
  - xlab_si.nuage_miq_automate
  tasks:
  - debug: msg="I am a playbook running as event callback and I am able to access nuage credentials"
  - debug: var=nuage_auth
  - debug: msg="So I'm now ready to do some work!"
```

miha-plesko:
Sorry for spamming, but gotta share this with you as well: so we'll put AE Class Attributes on NUAGE class (it's what this PR does). Then we'll instruct user to only set default values on Class (the username, the password etc.) - and all the underlying AE Instances will be able to access it 😎

^ if user was to set username/password on AE Instance then she would need to do it for every single instance. Whenever password would change, she would need to change it on all instances as well. But if we leverage AE Class default values, all instances can consume it so it needs to be only updated in one place i.e. default value needs to be updated on NUAGE class :)

PR #410, opened 2018-08-22
Do not check instance variables in specs for Transformation - Common - Utils

fdupont-redhat:
In previous code, we checked the instance variables which is not recommended and probably over-engineered. This PR relies only on method return. In the process, we removed the `task_and_vms` method which might not be that useful in real use.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1600870


fdupont-redhat:
@miq-bot add-label transformation, bug, gaprindashvili/yes, blocker
@miq-bot add-reviewer mkanoor

simaishi:

PR #413, opened 2018-08-27
Updated requested method to fix vm_reconfigure via rest-api.

billfitzgerald0120:
Issuing a vm_reconfigure disk_remove via rest-api was failing.
Very minor change (:disk_name was changed to 'disk_name')

```disk_num = disk['disk_name'].match(/_(\d).vmdk/)"```

https://bugzilla.redhat.com/show_bug.cgi?id=1620161

The rest-api call for vm_reconfigure uses the ```options[:disk_remove]``` hash which contains a string key of ```'disk_name'```.  The quota requested method failed since it was expecting disk_name as a symbol. 

The UI also uses the ```options[:disk_remove]``` hash with a string key of ```'disk_name'```, which doesn't fail the same way in quota because the hash type is HashWithIndifferentAccess. So, the method wasn't able to find the correct value (nil), but didn't fail. Although this ticket is specifically about quota, the actual vm_reconfigure works when initiated by the UI, but does not work properly when initiated by the rest-api.
Will open a new ticket to track the vm_configure rest-api issue.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/requested.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/requested.rb
@@ -158,7 +158,7 @@ def requested_storage(args_hash)
     end
     if args_hash[:resource].options[:disk_remove]
       args_hash[:resource].options[:disk_remove].each do |disk|
-        disk_num = disk[:disk_name].match(/_(\d).vmdk/)
+        disk_num = disk['disk_name'].match(/_(\d).vmdk/)
         next unless disk_num
         $evm.log(:info, "Reconfigure Disk Removal: #{disk.inspect}")
         disk_n_number = "disk_#{disk_num[1].succ}_size"
--- a/spec/automation/unit/method_validation/requested_spec.rb
+++ b/spec/automation/unit/method_validation/requested_spec.rb
@@ -240,6 +240,15 @@ describe "Quota Validation" do
       check_results(ws.root['quota_requested'], 10.megabytes, 2, 1, 4096.megabytes)
     end
 
+    it "removes a disk " do
+      setup_model("vmware_reconfigure")
+      @reconfigure_request.update_attributes(:options => {:src_ids => [@vm_vmware.id], :request_type => :vm_reconfigure,\
+      :disk_remove => [{"disk_name" => "freds disk", "persistent" => true, "thin_provisioned" => false,\
+      "dependent" => true, "bootable" => false}]})
+      ws = run_automate_method(reconfigure_attrs)
+      check_results(ws.root['quota_requested'], 0, 0, 1, 0.megabytes)
+    end
+
     it "minus 1 cpu and minus 2048 memory" do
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by billfitzgerald0120** [comment]
@miq-bot add_label bug

File: content/automate/ManageIQ/System/CommonMethods/QuotaMethods.class/__methods__/requested.rb [hunk 1]
```diff
@@ -158,7 +158,7 @@ def requested_storage(args_hash)
     end
     if args_hash[:resource].options[:disk_remove]
       args_hash[:resource].options[:disk_remove].each do |disk|
-        disk_num = disk[:disk_name].match(/_(\d).vmdk/)
+        disk_num = disk['disk_name'].match(/_(\d).vmdk/)
```
mkanoor: @billfitzgerald0120 Is this a symbol versus string issue, has it worked before with a symbol?
Do we have spec for this.

billfitzgerald0120: @mkanoor @tinaafitz  I re-created on 5.8.4.5 using the rest-api.  The UI worked with the original code but not when using the rest-api.   I tested this on master with the same results.  There is a requested_spec but it doesn't test the disk remove.

mkanoor: @billfitzgerald0120 Is the spec file missing from this PR?

billfitzgerald0120: @mkanoor I just added the spec file.

billfitzgerald0120:
@mkanoor @gmcculloug @tinaafitz The requested method was asking for a symbol.  The rest-api uses a hash which caused the requested method to fail and create an Error.  
The UI uses a HashWithIndifferentAccess and the method wasn't able to find the correct value (nil) but didn't fail and the disk was removed.

I change the requested method to use a string and the requested method works for both the UI and the rest-api, but the rest-api fails while trying to delete the disk.  We will need to fix this in a different PR.

billfitzgerald0120:
We tried to use get_option but that returned an array and we need a hash.

JPrause:
@miq-bot add_label blocker

JPrause:
@billfitzgerald0120 if this can be backported, can you add the gaprindashvili/yes label.

billfitzgerald0120:
@miq-bot add_label gaprindashvili/yes

simaishi:

```
$ git log -1
commit aeea686734529869a3ed7b94a8da06da05933d85
Date:   Tue Sep 4 17:29:19 2018 -0400

    Merge pull request #413 from billfitzgerald0120/requested_rest_api_tweek
    
    Updated requested method to fix vm_reconfigure via rest-api.
    
    (cherry picked from commit 758b168e4520da52a310bb50666330fae697d2c6)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1629096
```

PR #414, opened 2018-08-28
Added log_and_exit utility method

mkanoor:
This utility method allows a user to log and exit with a single method call.
It also tests the exit call from a spec

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
@@ -8,6 +8,16 @@ module ManageIQ
       module CommonMethods
         module Utils
           class LogObject
+            # If you want to log a message and exit without specifying a handle using the global $evm
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_exit(msg, code)
+            #
+            # If you want to log a message and exit using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_exit(msg, code, handle)
+            #
+            def self.log_and_exit(msg, exit_code, handle = $evm)
+              handle.log('info', "Script ending #{msg} code : #{exit_code}")
+              exit(exit_code)
+            end
             # If you want to log the root MiqAeObject and use the global $evm
             #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.root
             #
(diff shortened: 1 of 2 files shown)


mkanoor:
@bdunne 
Please review this is a tricky exit call in an Automate Method, which we have refrained from. I remember you doing some research on this in the past.

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -8,6 +8,16 @@ module System
...
           class LogObject
+            # If you want to log a message and exit without specifying a handle using the global $evm
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_exit(msg, code)
+            #
+            # If you want to log a message and exit using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_exit(msg, code, handle)
+            #
+            def self.log_and_exit(msg, exit_code, handle = $evm)
+              handle.log('info', "Script ending #{msg} code : #{exit_code}")
+              exit(exit_code)
```
gmcculloug: @billfitzgerald0120 @mkanoor Is the goal of this method to replace call in our existing model, like:

[ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job/wait_for_completion.rb#L48](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/wait_for_completion.rb#L48)

and

[ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM/update_provision_status.rb#L9](https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/VMProvision_VM.class/__methods__/update_provision_status.rb#L9)

mkanoor: @gmcculloug we recently fixed a bug in how MIQ_STOP can be used to a state machine which requires us to use the exit system call. Till now we have refrained from using exits in the newer methods because that causes the spec to exit. This is more a way to ensure we can test methods with exit in them.

bdunne:
@mkanoor It looks like you have a "Fix rubocop" commit, can you squash that?

bdunne:
I forget why we need to call `exit` rather than sending the "exit code" across the DRB connection and letting the script end

mkanoor:
@bdunne These are external customer written scripts that call the exit at many different points in their code, the exit(rc) naturally flows back thru the exit status
https://github.com/ManageIQ/manageiq-automation_engine/blob/608d372004d7b50b116ab0fb375a0934cc2f1f69/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb#L199

exit is a normal way for customers to end their scripts
If we have to teach them to do $evm.exit_code(8) that would be cumbersome.

bdunne:
Okay, well the post script will re-raise SystemExit, so I don't see any problems

PR #418, opened 2018-08-31
Transformation throttling - Base components

fdupont-redhat:
This PR is the first part of the throttling mechanism. Instead of using a transformation task as the scheduler / throttler, we use a dedicated automation request/task. The throttler creation logic is in a method called in `on_entry`, separately from WeightedUpdateStatus. This acts as a watchdog and creates a new throttler if needed. The throttler is a generic automation task with it's own state machine, so instantiating it is be done using `$evm.execute(:create_automation_request)`.

The state machine associated to the throttler task loops until no transformation task is left, or reaching its TTL. It is be responsible to adjust the limits of all the running tasks, but this PR doesn't implement the logic. The conversion host selection logic has also been moved to this task as it would solves the collisions issue. The idea of the TTL is to prevent never ending tasks: the throttler reads the existing state from the transformation task, so the next one can take over with no loss.

To ensure it acts as a *singleton*, We check in the throttler code that no other throttler is alive, i.e. `request_state == 'active'. If more than one is alive, only the eldest (`created_on`) continues, the others stop.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1594196

### DIFF
--- a/content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
+++ b/content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
@@ -5,27 +5,14 @@ module ManageIQ
         class AcquireTransformationHost
           def initialize(handle = $evm)
             @handle = handle
+            @task = ManageIQ::Automate::Transformation::Common::Utils.task(@handle)
           end
 
           def main
-            factory_config = @handle.get_state_var(:factory_config)
-            raise "No factory config found. Aborting." if factory_config.nil?
-
-            task = @handle.root['service_template_transformation_plan_task']
-            @handle.log(:info, "Task: #{task.inspect}")
-
-            transformation_host_type, transformation_host, transformation_method = ManageIQ::Automate::Transformation::TransformationHosts::Common::Utils.get_transformation_host(task, factory_config)
-            if transformation_host.nil?
-              @handle.log(:info, "No transformation host available. Retrying.")
+            if @task.get_option(:transformation_host_id).nil?
               @handle.root['ae_result'] = 'retry'
               @handle.root['ae_retry_server_affinity'] = true
-              @handle.root['ae_retry_interval'] = $evm.object['check_convert_interval'] || '1.minutes'
-            else
-              @handle.log(:info, "Transformation Host: #{transformation_host.name}.")
-              task.set_option(:transformation_host_id, transformation_host.id)
-              task.set_option(:transformation_host_name, transformation_host.name)
-              task.set_option(:transformation_host_type, transformation_host_type)
-              task.set_option(:transformation_method, transformation_method)
+              @handle.root['ae_retry_interval'] = 15.seconds
             end
           rescue => e
             @handle.set_state_var(:ae_state_progress, 'message' => e.message)
--- a/content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/posttransform.yaml
+++ b/content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/posttransform.yaml
@@ -10,24 +10,24 @@ object:
   fields:
   - State2:
       value: "/Transformation/Infrastructure/VM/${state_var#destination_ems_type}/SetDescription"
-      on_entry: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
-        => "Update description of VM", task_message => "Migrating")
+      on_entry: /Transformation/TransformationThrottler.Watch ; /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight
+        => 5, description => "Update description of VM", task_message => "Migrating")
       on_exit: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
         => "Update description of VM", task_message => "Migrating")
       on_error: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
         => "Update description of VM", task_message => "Migrating")
   - State5:
       value: "/Transformation/Infrastructure/VM/Common/PowerOn"
-      on_entry: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
-        => "Power-on VM", task_message => "Migrating")
+      on_entry: /Transformation/TransformationThrottler.Watch ; /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight
+        => 5, description => "Power-on VM", task_message => "Migrating")
       on_exit: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
         => "Power-on VM", task_message => "Migrating")
       on_error: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
         => "Power-on VM", task_message => "Migrating")
   - State8:
       value: "/Transformation/Infrastructure/VM/Common/CheckPoweredOn"
-      on_entry: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 40, description
-        => "Power-on VM", task_message => "Migrating")
+      on_entry: /Transformation/TransformationThrottler.Watch ; /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight
+        => 40, description => "Power-on VM", task_message => "Migrating")
       on_exit: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 40, description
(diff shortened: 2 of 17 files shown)


fdupont-redhat:

fdupont-redhat:
@miq-bot add-reviewer mkanoor

mkanoor:
@fdupont-redhat @gmcculloug 
I think we were planning on implementing this in the core as opposed to automate, the hoops that automate to do throttling is painful, we understand it can be done but that is not an ideal implementation. One idea we had is that Queue should have some controls on how many specific tasks  it can be run and make sure they run to completion or error before it dispatches another task of that kind. There would be a UI element to control how many tasks of a specific kind can be run at one time.

agrare:
@mkanoor I agree that this should not be in automate, but I really would steer away from embedding any more logic into MiqQueue /cc @kbrock

spec/content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+
+  let(:current_object) { Spec::Support::MiqAeMockObject.new }
+  let(:ae_service) do
+    Spec::Support::MiqAeMockService.new(root).tap do |service|
+      current_object.parent = root
+      service.current_object = current_object
+    end
+  end
+
+  before do
```
kbrock: fy:i this is `before(:each)` as well

content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
```diff
@@ -5,27 +5,14 @@ module Common
...
               @handle.root['ae_result'] = 'retry'
               @handle.root['ae_retry_server_affinity'] = true
-              @handle.root['ae_retry_interval'] = $evm.object['check_convert_interval'] || '1.minutes'
-            else
-              @handle.log(:info, "Transformation Host: #{transformation_host.name}.")
-              task.set_option(:transformation_host_id, transformation_host.id)
-              task.set_option(:transformation_host_name, transformation_host.name)
-              task.set_option(:transformation_host_type, transformation_host_type)
-              task.set_option(:transformation_method, transformation_method)
+              @handle.root['ae_retry_interval'] = 15.seconds
```
kbrock: does it make sense to have an overridable value `e.g. $evm.object['...'] || 15.seconds`
(honest question)

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/posttransform.yaml
```diff
@@ -10,49 +10,49 @@ object:
   fields:
   - State2:
       value: "/Transformation/Infrastructure/VM/${state_var#destination_ems_type}/SetDescription"
-      on_entry: /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight => 5, description
-        => "Update description of VM", task_message => "Migrating")
+      on_entry: /Transformation/TransformationThrottler.Watch ; /System/CommonMethods/MiqAe.WeightedUpdateStatus(weight
```
kbrock: stating my interpretation:
`TransformationThrottler.Watch` is basically an added state transformation to ensure there is back pressure on the producer

kbrock:
@mkanoor @agrare It does feel like this belongs out of automate and closer to the queue. That is the purpose of the queue construct: separate declaring the work and the parallelization effort.

It seems like controlling the number of workers in a pool would be best here - and not to modify the queue at all. Since this feature is coded directly in automate itself, it may be difficult to declare workers of this type.

Not sure if we can have a special purpose queue or pull some of this logic out of automate and make a full fledged service.

content/automate/ManageIQ/Transformation/Common.class/__methods__/acquiretransformationhost.rb
fdupont-redhat: Using `$evm.object['...']` means adding an attribute to the class. And the class is really generic. I'd defer that to another key/value store, such as `$evm.get_state_var('...')` or even data from the throttler request/task itself, as it sets the tempo for host allocation.

content/automate/ManageIQ/Transformation/StateMachines/VMTransformation.class/posttransform.yaml
fdupont-redhat: Yes, it checks that the throttler is still there and starts a new one if not. That's a watchdog, hence its code being really tight.

spec/content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils_spec.rb
fdupont-redhat: I'll fix it.

fdupont-redhat:
This feature doesn't belong to Automate and should be implemented in the backend. Closing this PR.

content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils.rb
```diff
@@ -0,0 +1,125 @@
...
+        class Utils
+          NAMESPACE = 'Transformation/StateMachines'.freeze
+          CLASS_NAME = 'TransformationThrottler'.freeze
+          DEFAULT_THROTTLER_ELECTION_POLICY = 'eldest_active'.freeze
+          DEFAULT_THROTTLER_TYPE = 'Default'.freeze
+          DEFAULT_THROTTLER_TTL = 3600
+          DEFAULT_TASKS_SCHEDULING_POLICY = 'fifo'.freeze
+          DEFAULT_LIMITS_ADJUSTMENT_POLICY = 'skip'.freeze
+
+          def self.log_and_raise(message, handle = $evm)
```
mkanoor: @fdupont-redhat 
@billfitzgerald0120 Has added this 
https://github.com/ManageIQ/manageiq-content/blob/a50ea6c95dfb2e2b51a69720e67f0fae973d1733/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb#L56
which could be used here

gmcculloug:
After discussing this issue we agreed to accept this current automate solution while the preferred backend code is being worked on to address the bigger concern of where this logic ultimately lives.

@fdupont-redhat will be working on this enhancement which will replace this code in a future version.

content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils.rb
fdupont-redhat: @mkanoor Good pointer. I'll give it a look for other parts of Transformation namespace, where I have similar needs. This throttling PR is a short term solution, so I don't think I'll do further changes, unless hitting bugs.

content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils.rb
```diff
@@ -0,0 +1,125 @@
...
+                :attrs         => { :ttl => throttler_ttl(handle) }
+              },
+              'admin',
+              true
+            )
+          end
+
+          def self.retry_or_die(handle = $evm)
+            return if current_throttler(handle).created_on.utc + throttler_ttl(handle) < Time.now.utc
+            return if active_transformation_tasks(handle).empty?
```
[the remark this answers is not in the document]
fdupont-redhat: @mkanoor The current election method is based on `created_on` attribute and elects the eldest alive request. I expect that if the eldest request will always be in the cache when younger ones will check. If the cache is not up-to-date, the worst case scenario is that the eldest request is still alive and will stop, leaving no throttler alive until the cache is refreshed, which is an acceptable situation.

spec/content/automate/ManageIQ/Transformation/TransformationThrottler.class/__methods__/utils_spec.rb
```diff
@@ -0,0 +1,291 @@
...
+  let(:current_object) { Spec::Support::MiqAeMockObject.new }
+  let(:ae_service) do
+    Spec::Support::MiqAeMockService.new(root).tap do |service|
+      current_object.parent = root
+      service.current_object = current_object
+    end
+  end
+
+  before(:each) do
+    ManageIQ::Automate::Transformation::TransformationThrottler::Utils.instance_variable_set(:@task, nil)
```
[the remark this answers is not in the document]
fdupont-redhat: Definitely. But fortunately, it doesn't have an impact :smile:

PR #420, opened 2018-09-04
Add storage profile support in vmware_best_fit_least_utilized method.

lfu:
- Refactor vmware_best_fit_least_utilized method.
- Add storage profile support.

Depends on https://github.com/ManageIQ/manageiq-automation_engine/pull/219.
https://bugzilla.redhat.com/show_bug.cgi?id=1617520

@miq-bot add_label bug, gaprindashvili/yes


mkanoor:
@lfu 
Can we resolve the allow_any_instance_of issue, you just have to stub out a find and return the exact object that you need. We have several examples in the spec folder
https://github.com/ManageIQ/manageiq-content/blob/master/spec/content/automate/ManageIQ/Transformation/Infrastructure/VM/Common.class/__methods__/restorevmattributes_spec.rb#L96


Also can you ensure that there is 100% coverage

You can do the following from your terminal

```
export CI=1
rspec <<yourspec_file>>
open coverage/index.html
```

If you can attach a screenshot of the coverage information it helps with the PR.

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,68 @@
...
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                @handle.log("info", "Args:    #{MIQ_ARGS.inspect}")
+                best_fit_least_utilized
+              end
+
+              def vm
```
mkanoor: @lfu 
A better style might be to use an instance variable like here
https://github.com/ManageIQ/manageiq-content/blob/758b168e4520da52a310bb50666330fae697d2c6/content/automate/ManageIQ/Transformation/Infrastructure/VM/Common.class/__methods__/restorevmattributes.rb#L26

This way you can just call vm whenever you need it and if it doesn't exist it will raise the error.

We should do the same for the prov method and object, we could actually call the method provision_request or just request. prov by itself is confusing if its provider or provision request

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,68 @@
...
+                prov = @handle.root["miq_provision"]
+                raise "VM not specified" if prov.vm_template.nil?
+                prov.vm_template
+              end
+
+              def best_fit_least_utilized
```
mkanoor: @lfu 
Can this function be broken into smaller functions not greater than 5 or 6 lines

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,68 @@
...
+            class VmwareBestFitLeastUtilized
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                @handle.log("info", "Args:    #{MIQ_ARGS.inspect}")
+                best_fit_least_utilized
+              end
+
```
mkanoor: @lfu 
Need to add private here, since only the main is exposed to the outside world

mkanoor:
@lfu you might be able to use Embedded Methods for logging, we have one defined here
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb

In the spec you would have to use something like this to stub out the embedded methods
https://github.com/ManageIQ/manageiq-content/pull/251

You need a require line for the embedded method in the spec
require File.join(ManageIQ::Content::Engine.root, 'content/automate/ManageIQ/System/CommonMethods/StateMachineMethods.class/__methods__/utility.rb')

and then you stub out the call
```ruby
 util_obj = instance_double("ManageIQ::Automate::System::CommonMethods::StateMachineMethods::Utility")
       # When the caller creates a new instance we send back the dummy instance
       allow(ManageIQ::Automate::System::CommonMethods::StateMachineMethods::Utility).to receive(:new).with(vm.name).and_return(util_obj)
       # When that instance calls normalize we stub it and pass back a canned response
       allow(util_obj).to receive(:normalize).with(no_args).and_return("blah")
```

lfu:
screen shot 2018-09-07 at 14 32 08

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,74 @@
...
+    module Infrastructure
+      module VM
+        module Provisioning
+          module Placement
+            class VmwareBestFitLeastUtilized
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def log_and_raise(message)
```
mkanoor: @lfu Shouldn't this be marked as private

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,74 @@
...
+
+              def main
+                @handle.log("info", "vm=[#{vm.name}], space required=[#{vm.provisioned_storage}]")
+                best_fit_least_utilized
+              end
+
+              private
+
+              def request
+                @request ||= @handle.root["miq_provision"].tap do |req|
```
mkanoor: @lfu 
You can have a spec that doesn't set the miq_provision attribute in the root object and call the method and it should raise an exception
Add another spec which has a request but no vm and you will be able to test the second log_and_raise

lfu:
screen shot 2018-09-10 at 11 31 01

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,74 @@
...
+
+                  s = storages.max_by(&:free_space)
+                  next if s.nil?
+                  host    = h
+                  storage = s
+                  min_registered_vms = h.vms.size
+                end
+
+                # Set host and storage
+                request.set_host(host) if host
```
mkanoor: @lfu What happens if host and storage are both nil we can't find any appropriate host or storage should we record that as an error and stop the process

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
```diff
@@ -1,37 +1,74 @@
...
+                request.set_storage(storage) if storage
+
+                @handle.log("info", "vm=[#{vm.name}] host=[#{host}] storage=[#{storage}]")
+              end
+
+              def log_and_raise(message)
```
mkanoor: @lfu 
Can you replace this method with this one
https://github.com/ManageIQ/manageiq-content/blob/a50ea6c95dfb2e2b51a69720e67f0fae973d1733/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb#L56

In the CFME Explorer you would have to add an embedded method pointing to this method and replace all calls to log_and_raise to use
```
ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_raise(message, @handle)
```

To help with the spec and embed methods look at this PR
https://github.com/ManageIQ/manageiq-content/pull/251/files

# This is the embedded method link, its only needed in the spec
 # Not in the actual file and ensures that the symbols are satisfied
 require File.join(ManageIQ::Content::Engine.root, 'content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb')






       # This is optional so that we dont have to the run the embedded method spec
       # This is a stub and returns canned responses with no computational overhead
       #
       allow(ManageIQ::Automate::System::CommonMethods::Utils::LogObject).to receive(:log_and_raise).with(msg, ae_service)

spec/content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized_spec.rb
```diff
@@ -20,6 +20,16 @@
   let(:root_object) { Spec::Support::MiqAeMockObject.new(:miq_provision => svc_miq_provision) }
   let(:ae_service)  { Spec::Support::MiqAeMockService.new(root_object) }
 
+  it 'requires miq_provision attribute in root object' do
+    new_service = Spec::Support::MiqAeMockService.new(Spec::Support::MiqAeMockObject.new)
+    expect { described_class.new(new_service).main }.to raise_error(RuntimeError, "ERROR - miq_provision object not passed in")
```
mkanoor: @lfu does it make sense that the message be checked using a regular expression
e.g.
```
expect { raise "oops" }.to raise_error(RuntimeError, /op/)
```
https://relishapp.com/rspec/rspec-expectations/v/2-11/docs/built-in-matchers/raise-error-matcher

This way the spec wont fail if someone where to make a slight change to the message

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
gmcculloug: The backend raises an error if the required resources have not been selected.  The best-fit methods do not have to deal with it.

See https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/provision/placement.rb#L13-L15

spec/content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized_spec.rb
mkanoor: @lfu 
Should the 2 error messages be consistent
"ERROR - miq_provision object not passed in"
"ERROR - VM not specified"



also is ERROR needed since this is going to logged as an error anyway in 

https://github.com/ManageIQ/manageiq-content/blob/a50ea6c95dfb2e2b51a69720e67f0fae973d1733/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb#L57

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
mkanoor: @gmcculloug Should there be a warning message in this method if we can't pick a host or a storage?

gmcculloug: That's fine if you want to add one.  The line immediately below this also logs if we have either object: `@handle.log("info", "vm=[#{vm.name}] host=[#{host}] storage=[#{storage}]")`

JPrause:
@miq-bot add_label blocker

mkanoor:
@lfu 
Try this for the log_and_raise example
``` ruby
it 'requires source vm in miq_provision' do
    miq_provision.update_attributes(:source => nil)
    expect(ManageIQ::Automate::System::CommonMethods::Utils::LogObject).to receive(:log_and_raise).once.with(/VM not specified/,ae_service).and_raise(RuntimeError)
    expect { described_class.new(ae_service).main }.to raise_error(RuntimeError)
  end 
```

content/automate/ManageIQ/Infrastructure/VM/Provisioning/Placement.class/__methods__/vmware_best_fit_least_utilized.rb
lfu: I don't think adding another log message should be necessary as there is already this line:
```
@handle.log("info", "vm=[#{vm.name}] host=[#{host}] storage=[#{storage}]")
```
at the end of this method, then later https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/provision/placement.rb#L13-L15 as @gmcculloug has pointed out.

mkanoor: @lfu @gmcculloug 
If there is no host or storage that message should atleast become a warning otherwise it will get lost in the log, if its a warning at least it would stand out. Can we change the level to warning if host or storage is nil

lfu: An error message would be thrown out if host or storage is nil.

```
    raise MiqException::MiqProvisionError, "Destination placement_ds_name not provided" if datastore.nil?
    raise MiqException::MiqProvisionError, "Destination placement_host_name and placement_cluster_name not provided" if host.nil? && cluster.nil?
```

mkanoor: @lfu 
That error message happens in a completely disconnected flow when the provision starts. The fact that the placement method didn't find a suitable host or storage is never reported and lost in a log message where the value is nil. As far as debugging is concerned we wont know that this method didn't find a suitable host or storage. Its important to report an error/warning when it first happens rather than have some other process down the road throw an error that it didn't get the required value.

gmcculloug: @mkanoor I would like to suggest that we address this issue in a separate PR or git issue as not to hold up the refactoring and storage profile fix this PR is trying to address.

But I wanted to note here, in my mind we have always allowed the backend to determine if all the requirements were met for the task, it did not fall on the automate script to know all the details of what the provider required.  And a single script may not provide all mandatory fields, they could be provided through another means.

For example, a user could use this script and potentially follow it with another placement script, as not to change our default methods, that uses different logic to provide the missing information.  Or the backend implementation could potentially change to determine missing fields differently based on new features or settings.

Having the automate method know these details does not see like the best place.

Again, I'm not against discussing this, but it feels like a different topic and probably should not hold up reviewing/merging this PR.

mkanoor:
@lfu 
If this needs to be back ported to G release the log embedded method doesn't exist in G release
https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/CommonMethods/Utils.class
This class only exists in master

lfu:
Will add another PR to implement the embedded log method once this one is merged.

simaishi:

```
$ git log -1
commit 85361f3a61f3d8e1ce5bdf2435b87712209c45c5
Date:   Thu Sep 20 15:57:05 2018 -0400

    Merge pull request #420 from lfu/storage_profile_1617520
    
    Add storage profile support in vmware_best_fit_least_utilized method.
    
    (cherry picked from commit 925e25be65f7badb183d9958fd2c64bafdcfc1b4)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1631557
```

PR #405, opened 2018-08-17
Change VMware URI to connect directly to ESXi

fdupont-redhat:
While investigating issues with large transformations in parallel, it was identified that vCenter is a bottleneck for VM transformation. The solution is to connect directly to the ESXi host. This PR provides the necessary changes to Automate, in order to connect to the ESXi host.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1618717

PR #421, opened 2018-09-13
Refactor and add specs for Transformation Hosts Common Utils

fdupont-redhat:
This PR adds specs for Transformation Hosts Common Utils method, according to https://github.com/ManageIQ/manageiq-content/issues/360. It also includes some changes to the utility methods for VMware to address comments from @jrafanie in https://github.com/ManageIQ/manageiq-content/pull/405 and make the Transformation Hosts Common Utils method more robust to IPv6.

Associated BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1600870


fdupont-redhat:
@miq-bot add-label transformation, enhancement, gaprindashvili/yes, test, blocker, technical debt

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -33,49 +30,77 @@ def self.transformation_hosts(ems, factory_config)
...
               send("virtv2vwrapper_options_#{task.get_option(:transformation_type)}_#{task.get_option(:transformation_method)}", task)
             end
 
-            def self.remote_command(task, transformation_host, command, stdin = nil, run_as = nil)
-              "ManageIQ::Automate::Transformation::TransformationHosts::#{task.get_option(:transformation_host_type)}::Utils".constantize.remote_command(transformation_host, command, stdin, run_as)
+            def self.vmware_uri_vddk(vm)
+              URI::Generic.build(
+                :scheme   => 'esx',
+                :userinfo => CGI.escape(vm.host.authentication_userid),
+                :host     => vm.host.ipaddress,
```
jrafanie: Looks good for IPv6 👍

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/utils.rb
```diff
@@ -33,49 +30,77 @@ def self.transformation_hosts(ems, factory_config)
...
+                :path     => '/',
+                :query    => { :no_verify => 1 }.to_query
```
jrafanie: What is the to_query method on hash?

fdupont-redhat: It converts this string to `?no_verify=1`, which convenient to escape input.

fdupont-redhat:
@miq-bot remove-label wip
@miq-bot add-reviewer mkanoor

content/automate/ManageIQ/Transformation/Infrastructure/VM/vmwarews.class/__methods__/utils.rb
```diff
@@ -7,53 +7,44 @@ module VMware
             class Utils
               require 'rbvmomi'
 
-              def initialize(handle = $evm)
-                @handle = handle
-              end
-
-              def main
+              def self.host_fingerprint(host)
```
mkanoor: @fdupont-redhat do you want to pass a handle into these class methods so that they can be tested

fdupont-redhat: @mkanoor do I need a handle to test them even though I don't use it in the method ?

mkanoor: @fdupont-redhat A missing handle means we are not logging any errors in the method we are expecting the exception to report an error. If you plan on writing specs it might be better to have a handle.

fdupont-redhat: @mkanoor 
I've added the handle but now Rubocop complains that I don't use it.
I probably miss something in the implementation of specs, as I don't see what this will bring. If my method raise, it will be caught by something like `expect(described_class.provider_connection).to raise(errormsg)`.

content/automate/ManageIQ/Transformation/Infrastructure/VM/vmwarews.class/__methods__/utils.rb
```diff
@@ -22,14 +22,14 @@ def self.host_fingerprint(host)
...
-              def self.provider_connection(ems)
+              def self.provider_connection(ems, handle = $evm)
                 @provider_connection ||= RbVmomi::VIM.connect(:host => ems.ipaddress || ems.hostname, :user => ems.authentication_userid, :password => ems.authentication_password, :insecure => true)
               rescue => e
                 raise "Could not connect to #{ems.name}: #{e.message}" if c.nil?
               end
               private_class_method :provider_connection
 
-              def self.vm_get_ref(vim, vm)
+              def self.vm_get_ref(vim, vm, handle = $evm)
```
mkanoor: @fdupont-redhat would you be logging these errors or just the raise an error. If you are going to be logging or using your shared log_and_raise the handle will get used. I am thinking these are all class methods

fdupont-redhat: @mkanoor These are indeed class methods, so I will log the errors from the catch all rescue in the calling method, eventually using shared log_and_raise.

content/automate/ManageIQ/Transformation/Infrastructure/VM/vmwarews.class/__methods__/utils.rb
```diff
@@ -7,53 +7,44 @@ module VMware
...
+                cert = OpenSSL::X509::Certificate.new(ssl_client.peer_cert)
+                ssl_client.sysclose
+                tcp_client.close
+
+                Digest::SHA1.hexdigest(cert.to_der).upcase.scan(/../).join(":")
               end
 
-              def main
+              def self.provider_connection(ems, handle = $evm)
+                @provider_connection ||= RbVmomi::VIM.connect(:host => ems.ipaddress || ems.hostname, :user => ems.authentication_userid, :password => ems.authentication_password, :insecure => true)
```
mkanoor: @fdupont-redhat should the secure true/false be configurable. This makes the assumption its always insecure

mkanoor: @fdupont-redhat why is the provider_connection an instance variable, this looks like a class method. This connection cannot have scope outside of the process that is using this, once the process ends the provider_connection is meaningless. Also we need to ensure that this provider connection doesn't get into a $evm.set_state_var.
So could the provider connection not be exposed to callers maybe we could pass in a block into this utility method and it would execute the block on that provider. Or have a shared class that actually has instance methods which can execute the code on the provider but keeps the connection private and doesn't allow other classes access to it.
Exposing the connection object this way can lead to misuse.

fdupont-redhat: @mkanoor To be honest, I would remove these methods: `provider_connection`, `vm_get_ref`, `vm_rename`. I think they should be in the model and only `vm_rename` should be exposed in `vm` class. This is currently unused as we don't rename the VMs after migration. I just kept them because it was already there.

simaishi:

simaishi:
@fdupont-redhat `hammer/yes` ?

simaishi:
@fdupont-redhat That's how I noticed this PR 😄  But didn't realize all files were deleted in the new PR...  Anyway, `hammer/yes` will help as https://github.com/ManageIQ/manageiq-content/pull/441 backport conflicts because of this PR.

Hammer backport details:

```
$ git log -1
commit ff0f2e10b527ea2ea4af84ac3841a58589d6342b
Date:   Thu Sep 27 09:37:05 2018 -0400

    Merge pull request #421 from fdupont-redhat/v2v_refactor_and_specs_transformation_host
    
    Refactor and add specs for Transformation Hosts Common Utils
    
    (cherry picked from commit fcc47328a0ea2551904d9f58105c4c8e5fdaffb8)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1600870
```

PR #423, opened 2018-09-17
Adding log_and_notify method into log_object embedded method.

billfitzgerald0120:
This utility method allows an user to create a notification and log message with a single method call.
You can create an info, warning, success or error notification.
You can create an info, warn or error log message.

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
@@ -1,6 +1,7 @@
 #
 # Description: Display Log messages for object attributes for root, current or desired object.
 # Added InspectMe functionality
+# Added log_and_notify
 
 module ManageIQ
   module Automate
@@ -8,6 +9,12 @@ module ManageIQ
       module CommonMethods
         module Utils
           class LogObject
+            NOTIFY_LEVEL_TO_LOG_LEVEL = {
+              'info'    => 'info',
+              'warning' => 'warn',
+              'error'   => 'error',
+              'success' => 'info'
+            }.freeze
             # If you want to log a message and exit without specifying a handle using the global $evm
             #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_exit(msg, code)
             #
@@ -18,6 +25,20 @@ module ManageIQ
               handle.log('info', "Script ending #{msg} code : #{exit_code}")
               exit(exit_code)
             end
+
+            # If you want to create a notification and log a message without specifying a handle using the global $evm
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(notify_level, msg, subject)
+            #
+            # If you want to create a notification and log a message using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(notify_level, msg, subject, handle)
+            #
+            # Valid types are : info, warning, error and success
+            def self.log_and_notify(notify_level, message, subject, handle = $evm)
+              raise "Invalid notify level #{notify_level}" unless NOTIFY_LEVEL_TO_LOG_LEVEL.keys.include?(notify_level.downcase.to_s)
+              handle.create_notification(:level => notify_level.downcase, :message => message, :subject => subject)
+              handle.log(NOTIFY_LEVEL_TO_LOG_LEVEL[notify_level.downcase.to_s], message)
+            end
+
             # If you want to log the root MiqAeObject and use the global $evm
             #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.root
             #
--- a/spec/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object_spec.rb
+++ b/spec/content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object_spec.rb
@@ -80,6 +80,61 @@ describe ManageIQ::Automate::System::CommonMethods::Utils::LogObject do
     end
   end
 
+  shared_examples_for "log_and_notify" do
+    it "check" do
+      expect(ae_service).to receive(:create_notification).with(:level => notify_level, :message => message, :subject => vm1)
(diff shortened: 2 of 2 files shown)


billfitzgerald0120:
@miq-bot add_reviewer @tinaafitz

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -18,6 +19,28 @@ def self.log_and_exit(msg, exit_code, handle = $evm)
...
+            #
+            # If you want to create a notification and log a message using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(:level, msg, subject, handle)
+            #
+            def self.log_and_notify(level, message, subject, handle = $evm)
+              handle.create_notification(:level => notify_level(level.downcase.to_s), :message => message, :subject => subject)
+              handle.log(log_level(level.downcase.to_s), message)
+            end
+
+            def self.log_level(level)
```
mkanoor: @billfitzgerald0120 What is this trying to do, does notify use warning instead of warn, if the caller is passing in level couldn't we enforce that the level be "info","warn",or"error"

billfitzgerald0120:
log_object_spec coverage

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
billfitzgerald0120: @mkanoor yes, notify uses warning and the log uses warn. The notify accepts info, warning, 
 error and success.  The log accepts info, warn and error.  For level, If a invalid value like 'fred' is passed in then it is changed to 'info' for both notify and log.

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -18,6 +19,28 @@ def self.log_and_exit(msg, exit_code, handle = $evm)
...
+              log_level = level == "warning" ? "warn" : level
+              %w(info warn error).include?(log_level) ? log_level : "info"
+            end
+
+            def self.notify_level(level)
```
mkanoor: @billfitzgerald0120 Instead of using a function can we just use a hash e.g
```ruby
NOTIFY_LEVEL_TO_LOG_LEVEL = { 
  'info'    => 'info',
  'warning' => 'warn',
  'error'   => 'error',
  'success' => 'info'
}
NOTIFY_LEVEL_TO_LOG_LEVEL.default = 'info'
```
To fetch values you could use something like
```ruby
puts NOTIFY_LEVEL_TO_LOG_LEVEL['fred']
puts NOTIFY_LEVEL_TO_LOG_LEVEL['warning']
```

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -18,6 +19,28 @@ def self.log_and_exit(msg, exit_code, handle = $evm)
...
               exit(exit_code)
             end
+
+            # If you want to create a notification and log a message without specifying a handle using the global $evm
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(:level, msg, subject)
+            #
+            # If you want to create a notification and log a message using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(:level, msg, subject, handle)
+            #
+            def self.log_and_notify(level, message, subject, handle = $evm)
```
mkanoor: @billfitzgerald0120 We need to decide if the level passed into this method is a notify level or a log_level.

When this function is called will the user pass in level to be one of these ("info","error","warn","debug") or would it pass in the notify level ("success","error","warning","info") We have to pick one.

I was thinking it would be the notify level and then we would map from the notify_level to the log_level
so when we call 
``` ruby
def self.log_and_notify(notify_level, message, subject, handle = $evm)
  raise "Invalid notify level #{notify_level}" unless NOTIFY_LEVEL_TO_LOG_LEVEL.keys.include?(notify_level.downcase)
  handle.create_notification(:level => notify_level.downcase, :message => message, :subject => subject)
  handle.log(NOTIFY_LOG_TO_LOG_LEVEL(notify_level.downcase), message)
end
```

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -18,6 +25,20 @@ def self.log_and_exit(msg, exit_code, handle = $evm)
               handle.log('info', "Script ending #{msg} code : #{exit_code}")
               exit(exit_code)
             end
+
+            # If you want to create a notification and log a message without specifying a handle using the global $evm
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(:level, msg, subject)
```
mkanoor: @billfitzgerald0120 Do you want to change the comment to say :notify_level instead of level

content/automate/ManageIQ/System/CommonMethods/Utils.class/__methods__/log_object.rb
```diff
@@ -18,6 +25,20 @@ def self.log_and_exit(msg, exit_code, handle = $evm)
...
+            #
+            # If you want to create a notification and log a message using a specific handle
+            #    ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_notify(:level, msg, subject, handle)
```
mkanoor: @billfitzgerald0120 Also here in the command the level should be notify level

billfitzgerald0120:
@mkanoor Changed comments as requested, please review

billfitzgerald0120:
Screenshot of method:

log_and_notify test method


Screenshot of simulation:

simulation - test log_and_notify embedded method

ganeshhubale:
@billfitzgerald0120 @mkanoor 
Is this RFE back-ported to 5.9z?

gmcculloug:
@ganeshhubale No, as an enhancement these changes normally would not be back-ported.  This is available in the `Hammer` release.

PR #429, opened 2018-09-21
Yamls to support backend-initiated notifications in v2v for Successful and Failed Requests

AparnaKarve:
Fixes https://github.com/ManageIQ/manageiq-v2v/issues/265
Fixes https://github.com/ManageIQ/manageiq-v2v/issues/578

Supporting Core PR - https://github.com/ManageIQ/manageiq/pull/18012

### DIFF
--- a/content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_failure.yaml
+++ b/content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_failure.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: request_completed_with_failure
+    inherits: 
+    description: 
+  fields:
+  - rel5:
+      value: "/System/Policy/request_completed_with_failure"
--- a/content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_success.yaml
+++ b/content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_success.yaml
@@ -1,0 +1,12 @@
+---
+object_type: instance
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: request_completed_with_success
+    inherits: 
+    description: 
+  fields:
+  - rel5:
+      value: "/System/Policy/request_completed_with_success"


content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_failure.yaml
```diff
@@ -0,0 +1,12 @@
...
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: request_completed_with_failure
+    inherits: 
+    description: 
+  fields:
+  - rel5:
+      value: "/System/Policy/request_completed_with_failure"
```
mkanoor: @AparnaKarve 
/System/Policy/request_completed_with_failure is not defined in the database in https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Policy.class are you planning on using the missing instance for this

content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_success.yaml
```diff
@@ -0,0 +1,12 @@
...
+version: 1.0
+object:
+  attributes:
+    display_name: 
+    name: request_completed_with_success
+    inherits: 
+    description: 
+  fields:
+  - rel5:
+      value: "/System/Policy/request_completed_with_success"
```
mkanoor: @AparnaKarve 
/System/Policy/request_completed_with_success is not defined in https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Policy.class are you planning on using the missing instance?

content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_failure.yaml
AparnaKarve: @mkanoor Should I also define request_completed_with_failure under https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Policy.class ?
The way I have it right now seems to work.

content/automate/ManageIQ/System/Event/RequestEvent/Request.class/request_completed_with_success.yaml
AparnaKarve: same question as above https://github.com/ManageIQ/manageiq-content/pull/429#discussion_r219965042

mkanoor:
@lfu do you have comments on this?

lfu:
This does not look right to me. Usually the request comes in to kick out automate actions defined in ```/System/Policy```. But these new requests don't have any actions defined in ```/System/Policy```.

Why these new request instances are needed for notification?
cc @gmcculloug

lfu:
@tinaafitz @billfitzgerald0120 How do we implement notification from backend?

AparnaKarve:
Is there a prior example in the existing codebase that comes close to my current requirement that I can refer to and follow along?

tinaafitz:
Hi @AparnaKarve @lfu You can create notifications without using Automate as is shown here:
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_server.rb#L373

Can you use that approach, or do you need to have it setup in Automate content?
Automate uses some predefined generic notification types. It seems like your notifications are specific, and should be added to the notification_types table.

Can we get together on Monday to help us better understand the requirements?
/cc @mkanoor

AparnaKarve:
@tinaafitz`Notification.create` was exactly what I tried before I came down this route.
I have no idea why, but it does not seem to work for `MiqRequest`

Then I started looking up some of the existing `MiqRequest` notifications that *do* work, like `request_approved` (shown below) 
screen shot 2018-09-28 at 10 55 27 am

What I noticed was, `request_approved` and `request_denied` type of notifications were using `call_automate_event_queue`
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_request.rb#L254
https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_request.rb#L270

which is what I have used in ManageIQ/manageiq#18012
(this is the supporting MIQ core PR for the current PR)

I believe my use case is very similar to the above two (`request_approved` and `request_denied`), except that I'm targeting MiqRequests of the type `ServiceTemplateTransformationPlanRequest`, which are v2v specific

Let's get together on Monday and we can discuss this more.
Thanks Tina!

AparnaKarve:
As mentioned above `call_automate_event_queue` seems like the only way to go to create Notifications from `MiqRequest`, @tinaafitz @lfu let me know if you have other suggestions.

tinaafitz:
@AparnaKarve I'm wondering if it's possible that the ```Notification.create``` didn't work properly initially because there wasn't a notification type defined for your new notification.

AparnaKarve:
@tinaafitz That could very well be the case that the new Notification type was somehow not found when the `Notification.create` call was made.

This is something really strange --  even though the new Notification type exists in the `notification_types` table, the `Notification.create` call fails from the app and also from rails console. Any idea why that could be?

AparnaKarve:
@tinaafitz I may have identified the issue with `Notification.create` ...

It looks like in notification_types.yml, when I set the audience to either `user` or `group`, `Notification.create` does not seem to work.

In Rails console, I get the following error if I use a notification_type that has it's audience set to user/group
```
> Notification.create(:type => "generic_task_finish", :options => {:message => "test"})
  NotificationType Load (0.5ms)  SELECT  "notification_types".* FROM "notification_types" WHERE "notification_types"."name" = $1 LIMIT $2  [["name", "generic_task_finish"], ["LIMIT", 1]]
  NotificationType Inst Including Associations (0.1ms - 1rows)
   (0.2ms)  BEGIN
   (0.1ms)  ROLLBACK
NoMethodError: undefined method `id' for nil:NilClass
	from ~/manageiq/app/models/notification_type.rb:22:in `subscriber_ids'
	from ~/manageiq/app/models/notification.rb:65:in `set_notification_recipients'
	from /Users/akarve/.rvm/gems/ruby-2.4.1/gems/activesupport-5.0.7/lib/active_support/callbacks.rb:382:in `block in make_lambda'
	from /Users/akarve/.rvm/gems/ruby-2.4.1/gems/activesupport-5.0.7/lib/active_support/callbacks.rb:169:in `block (2 levels) in halting'
```

--------

So I have now changed the audience of my new notification_types to `global` , which seems to work.
```
:audience: global
```

With the changes made in ManageIQ/manageiq#18012, it looks like we no longer need this PR.

Thank you for your tips -- really appreciated!!

PR #433, opened 2018-09-27
Fix virt-v2v kill

fdupont-redhat:
In the course of a discussion with Richard W.M. Jones, he said that:

> It's not too nice to send kill -9 to virt-v2v because it means none of the at-exit handlers get to run, so it will leave temporary files all over the place.  It's better to send an ordinary kill signal (eg. SIGTERM).  If virt-v2v doesn't exit after some grace period, eg. 30 seconds, then it's a bug, but maybe you could then send SIGKILL.

This PR aims at fixing this by first using SIGTERM to kill virt-v2v with a grace period of 30 seconds, using retry. After that grace period, virt-v2v is sent SIGKILL, as it was the case before.

While I was at it, I added the specs for this method.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1633526

### DIFF
--- a/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v.rb
+++ b/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v.rb
@@ -1,28 +1,38 @@
 module ManageIQ
   module Automate
     module Transformation
-      module TransformationHost
+      module TransformationHosts
         module Common
           class KillVirtV2V
             def initialize(handle = $evm)
               @handle = handle
+              @task = ManageIQ::Automate::Transformation::Common::Utils.task(@handle)
             end
 
-            def task_virtv2v_state(task, transformation_host)
+            def task_virtv2v_state(transformation_host)
               require 'json'
-              return if task.get_option(:virtv2v_started_on).blank? || task.get_option(:virtv2v_finished_on).present?
-              return if task.get_option(:virtv2v_wrapper).blank?
-              result = Transformation::TransformationHosts::Common::Utils.remote_command(task, transformation_host, "cat '#{task.get_option(:virtv2v_wrapper)['state_file']}'")
+              return if @task.get_option(:virtv2v_started_on).blank? || @task.get_option(:virtv2v_finished_on).present?
+              return if @task.get_option(:virtv2v_wrapper).blank?
+              result = ManageIQ::Automate::Transformation::TransformationHosts::Common::Utils.remote_command(@task, transformation_host, "cat '#{@task.get_option(:virtv2v_wrapper)['state_file']}'")
               return if !result[:success] || result[:stdout].empty?
               JSON.parse(result[:stdout])
             end
 
+            def kill_virtv2v(transformation_host, pid)
+              signal = 'KILL'
+              unless @handle.get_state_var('virtv2v_graceful_kill')
+                signal = 'TERM'
+                @handle.set_state_var('virtv2v_graceful_kill', true)
+                @handle.root['ae_result'] = 'retry'
+                @handle.root['ae_retry_interval'] = '30.seconds'
+              end
+              ManageIQ::Automate::Transformation::TransformationHosts::Common::Utils.remote_command(@task, transformation_host, "kill -s #{signal} #{pid}")
+            end
+
             def main
-              task = @handle.vmdb(:service_template_transformation_plan_task).find_by(:id => @handle.root['service_template_transformation_plan_task_id'])
-              transformation_host = @handle.vmdb(:host).find_by(:id => task.get_option(:transformation_host_id))
-              virtv2v_state = task_virtv2v_state(task, transformation_host)
-              @handle.log(:info, "VirtV2V State: #{virtv2v_state.inspect}")
-              Transformation::TransformationHosts::Common::Utils.remote_command(task, transformation_host, "kill -9 #{virtv2v_state['pid']}") if virtv2v_state.present?
+              transformation_host = @handle.vmdb(:host).find_by(:id => @task.get_option(:transformation_host_id))
+              virtv2v_state = task_virtv2v_state(transformation_host)
+              kill_virtv2v(transformation_host, virtv2v_state['pid']) if virtv2v_state.present?
             rescue => e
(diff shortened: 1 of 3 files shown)


fdupont-redhat:
@miq-bot add-label transformation, bug, technical debt, hammer/yes

spec/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+  end
+
+  context "#task_virtv2v_state" do
+    it "when virtv2v not started" do
+      allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(nil)
+      expect(described_class.new(ae_service).task_virtv2v_state(svc_model_host)).to be_nil
+    end
+
+    it "when virtv2v is finished" do
+      allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(Time.now.utc - 1)
```
gmcculloug: @fdupont-redhat There is a lot of shared setup between these tests.  I recommend creating context so you can share the setup between tests.

For example, 
The `allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(Time.now.utc - 1)` could define one context.  with several tests.  

Inside of that context would be another that does a before block for `allow(svc_model_task).to receive(:get_option).with(:virtv2v_finished_on).and_return(nil)` and list the tests under it.

Can you take a shot at reorganizing these and let me know if you need help.

spec/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+  before do
+    allow(ManageIQ::Automate::Transformation::Common::Utils).to receive(:task).with(ae_service).and_return(svc_model_task)
+    allow(svc_model_task).to receive(:get_option).with(:transformation_host_id).and_return(svc_model_host.id)
+
+    allow(ae_service).to receive(:vmdb).with(:host).and_return(svc_vmdb_handle_host)
+    allow(svc_vmdb_handle_host).to receive(:find_by).with(:id => svc_model_host.id).and_return(svc_model_host)
+  end
+
+  before(:each) do
+    ManageIQ::Automate::Transformation::TransformationHosts::Common::KillVirtV2V.instance_variable_set(:@task, nil)
```
gmcculloug: `before(:each)` is the default action so it is the same as the `before` block above.  This line can just move up in to the previous block.

spec/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    context "when virtv2v is not started" do
+      before { allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(nil) }
+      it { expect(described_class.new(ae_service).task_virtv2v_state(svc_model_host)).to be_nil }
+    end
+
+    context "when virtv2v is started" do
+      before { allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(Time.now.utc - 1) }
+
+      context "when virtv2v is finished" do
+        before { allow(svc_model_task).to receive(:get_option).with(:virtv2v_finished_on).and_return(Time.now.utc) }
```
gmcculloug: Same, move the `allow` to the test block.

spec/content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+    allow(svc_vmdb_handle_host).to receive(:find_by).with(:id => svc_model_host.id).and_return(svc_model_host)
+  end
+
+  before(:each) do
+    ManageIQ::Automate::Transformation::TransformationHosts::Common::KillVirtV2V.instance_variable_set(:@task, nil)
+  end
+
+  context "#task_virtv2v_state" do
+    context "when virtv2v is not started" do
+      before { allow(svc_model_task).to receive(:get_option).with(:virtv2v_started_on).and_return(nil) }
```
gmcculloug: You don't need the `before` block if there is only one test.  Just include the `allow` line in the test block.

content/automate/ManageIQ/Transformation/TransformationHosts/Common.class/__methods__/killvirtv2v.rb
```diff
@@ -1,28 +1,38 @@
...
 
+            def kill_virtv2v(transformation_host, pid)
+              signal = 'KILL'
+              unless @handle.get_state_var('virtv2v_graceful_kill')
+                signal = 'TERM'
+                @handle.set_state_var('virtv2v_graceful_kill', true)
+                @handle.root['ae_result'] = 'retry'
+                @handle.root['ae_retry_interval'] = '30.seconds'
+              end
+              ManageIQ::Automate::Transformation::TransformationHosts::Common::Utils.remote_command(@task, transformation_host, "kill -s #{signal} #{pid}")
```
gmcculloug: I was thinking about how to write this without using `unless` since we normally only use that for simple one-line conditions.

My suggestion would be to move the `kill_signal` logic into a separate method.

```ruby
def kill_virtv2v(transformation_host, pid)
  ManageIQ::Automate::Transformation::TransformationHosts::Common::Utils.remote_command(@task, transformation_host, "kill -s #{kill_signal} #{pid}")
end

def kill_signal
  if @handle.get_state_var('virtv2v_graceful_kill')
    'KILL'
  else
    @handle.set_state_var('virtv2v_graceful_kill', true)
    @handle.root['ae_retry_interval'] = '30.seconds'
    'TERM'
  end
end
```

fdupont-redhat:
@gmcculloug Do you want more changes ?

fdupont-redhat:
This change has been implemented in https://github.com/ManageIQ/manageiq-content/pull/441. Closing this PR.

PR #391, opened 2018-08-02
Transformation - Refactor task and vm collection

fdupont-redhat:
In [PR#382](https://github.com/ManageIQ/manageiq-content/pull/382), @gmcculloug proposed to refactor the way we collect the task and vm at the beginning of our method. This PR provides a implementation of this proposal, as a utility method that can be embedded. This will allow refactoring all the Transformation code to use it.

PR #440, opened 2018-10-06
[GAPRINDASHVILI] Add transformation utils methods

gmcculloug:
Missing util methods for v2v for Gaprindashvili branch.  Should get tests green.

@simaishi This is the same exact same changes as https://github.com/ManageIQ/manageiq-content/pull/391 just with squashed commits.  Ultimately we probably just want to backport the other PR and close this.


simaishi:
Backported https://github.com/ManageIQ/manageiq-content/pull/391

PR #458, opened 2018-10-29
[WIP] Set Ansible job limit to IP address instead of name.

tinaafitz:
Use IP address instead of name for target since VM name can contain spaces.

fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1643467

### DIFF
--- a/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb
+++ b/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb
@@ -30,7 +30,7 @@ module ManageIQ
 
                 def target
                   vm = @handle.root['vm'] || vm_from_request
-                  vm.name if vm
+                  vm.try(:floating_ip_addresses).try(:first) || vm.try(:ipaddresses).try(:first) if vm
                 end
 
                 def vm_from_request
--- a/spec/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job_spec.rb
+++ b/spec/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job_spec.rb
@@ -54,7 +54,8 @@ describe ManageIQ::Automate::AutomationManagement::AnsibleTower::Operations::Sta
     current_object = Spec::Support::MiqAeMockObject.new(:job_template_name => job_template.name)
     current_object.parent = root_object
     service.object = current_object
-    job_args[:limit] = vm.name
+    job_args[:limit] = ip_addr
+    allow(svc_vm).to receive(:ipaddresses).and_return([ip_addr])
     expect(job_class).to receive(:create_job).with(anything, job_args).and_return(svc_job)
     described_class.new(service).main
     expect(service.get_state_var(:ansible_job_id)).to eq(job.id)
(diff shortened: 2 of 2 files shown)


billfitzgerald0120:
@tinaafitz looks good

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb
```diff
@@ -30,7 +30,7 @@ def main
 
                 def target
                   vm = @handle.root['vm'] || vm_from_request
-                  vm.name if vm
+                  vm.try(:floating_ip) || vm.try(:ipaddresses).first if vm
```
gmcculloug: This line is not doing what you think it is.  The property `floating_ip` or even `floating_ips` is returning an `FloatingIP` instance which you would then have to call `.address` on to get the actual IP address.
See https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/vm.rb#L11-L14

Instead you want to reference the virtual columns `ipaddresses` and `floating_ip_addresses` defined here which are arrays of IP address strings: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/vm.rb#L32-L33

I would suggest you follow the logic already defined here which returns the IP address string to use:
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb#L45-L52

tinaafitz: @gmcculloug Thanks for bringing that to my attention.  Please let me know if there is anything else you would like changed.

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb
```diff
@@ -30,7 +30,7 @@ def main
 
                 def target
                   vm = @handle.root['vm'] || vm_from_request
-                  vm.name if vm
+                  vm.try(:floating_ip_addresses).try(:first) || vm.try(:ipaddresses).try(:first) if vm
```
gmcculloug: We also want to keep the `raise` logic from the other code sample I pointed.  If we do not have a VM for some reason, or the IP address is empty, this code will call Ansible without a limit which would try to run against all the inventory defined on the target JobTemplate.

https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb#L45-L52

mkanoor: @tinaafitz 
The vm.name was used for running job templates on freshly MiQ provisioned VMs and the only way we can have the tower update its inventory and run the job is with the name and not the IP address. Since this is a method I think we should allow users to control if we should use the name or the IP address.

phospi:
Hi there,

it's definitely correct that a limit is very important to avoid running a job template against the complete inventory. But you can as well shift the responsibility to the manageiq adminstrator by passing a limit via service dialog. 

Did you keep in mind that the ansible "hostname" identifier string must exactly match with the ansible "limit" parameter? If the ansible dynamic inventory sync pulls the node names from cloudforms and later on the limit is set as ip adress then this is not going to work as far as my tests showed.

Kind Regards,
Philipp

tinaafitz:
Thanks @mkanoor @phospi for providing additional information. I changed this PR to WIP while we figure out a solution.

mkanoor:
@tinaafitz @phospi 

When this method was implemented it was with the sole intention of running an Ansible JobTemplate as a post provision method.

As we have add more integration with Ansible we have been using other ways of getting the host as is evident in this method
https://github.com/ManageIQ/manageiq-content/blob/166a3df922e31953f63db76dd644df734fd1a40d/content/automate/ManageIQ/System/Request.class/__methods__/order_ansible_playbook.rb#L37

Maybe the launch_ansible_job method can be enhanced to look for a provided limit (hostname) first before delving into the vmname.

https://github.com/ManageIQ/manageiq-content/blob/166a3df922e31953f63db76dd644df734fd1a40d/content/automate/ManageIQ/AutomationManagement/AnsibleTower/Operations/StateMachines/Job.class/__methods__/launch_ansible_job.rb#L116

JPrause:
@tinaafitz if this will be able to be backported, can you add the hammer/yes flag.

phospi:
just for the record. We edited ansible cloudforms.py and ManageIQ launch_ansible_job.rb that every space is replaced by an underscore. This works very well for us.

PR #470, opened 2018-11-12
V2V - Workaround for network mappings storage

fdupont-redhat:
When migrating a VM to OpenStack, we need the IP addresses of the source virutal machine to create the associated network ports. This PR add a statement to store the network mappings, which contain the IP addresses if VMware tools are running, into the task options hash for later consumption.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1649020

### DIFF
--- a/content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
+++ b/content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
@@ -25,6 +25,8 @@ module ManageIQ
 
           def main
             raise 'Preflight check has failed' unless @task.preflight_check
+            # TODO(fdupont-redhat): Find why network mappings are not stored
+            # by call to preflight_check.
             @task.set_option(:network_mappings, @task.network_mappings)
             %w(task_options factory_config).each { |ci| send("populate_#{ci}") }
           rescue => e
--- a/content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
+++ b/content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
@@ -25,6 +25,7 @@ module ManageIQ
 
           def main
             raise 'Preflight check has failed' unless @task.preflight_check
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@miq-bot add-label transformation, bug, hammer/yes, blocker

content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
```diff
@@ -25,6 +25,7 @@ def populate_factory_config
 
           def main
             raise 'Preflight check has failed' unless @task.preflight_check
+            @task.set_option(:network_mappings, @task.network_mappings)
```
gmcculloug: @fdupont-redhat Seems odd to take the method data from a task and put it "into the task options hash for later consumption".  Is this so it can be used through the API later?  If so it would be better to expose it there then to store the data that the object already generates.

cc @roliveri @agrare

fdupont-redhat:
It's a workaround until I find why it's not set by the call to
`preflight_check`. Without it, migration fails and blocks QE.

Le lun. 12 nov. 2018 à 22:43, Greg McCullough <notifications@github.com> a
écrit :

content/automate/ManageIQ/Transformation/Common.class/__methods__/assesstransformation.rb
fdupont-redhat: It's a workaround until I find why it's not set by the call to `preflight_check`. Without it, migration fails and blocks QE.

gmcculloug: @fdupont-redhat I need to understand what this really means:
> why it's not set by the call to preflight_check` 

I suggest we focus on this a little before jumping right to this work-around.  We can discuss.

fdupont-redhat:
@gmcculloug found the solution to preflight_check not setting the network mappings: it was not saved in the DB. Another PR fixes the issue: https://github.com/ManageIQ/manageiq/pull/18194, so this one can be closed.

PR #476, opened 2018-11-21
Send the service id to Tower, by default, as an extra var.

h4wkmoon:


content/automate/ManageIQ/AutomationManagement/AnsibleTower/Service/Provisioning/StateMachines/Provision.class/__methods__/preprovision.rb
```diff
@@ -46,8 +46,9 @@ def examine_request(service)
                   def modify_job_options(service)
                     # Example how to programmatically modify job options:
                     job_options = service.job_options
-                    job_options[:limit] = 'someHost'
-                    job_options[:extra_vars]['flavor'] = 'm1.small'
+                    #job_options[:limit] = 'someHost'
+                    #job_options[:extra_vars]['flavor'] = 'm1.small'
+                    job_options[:extra_vars]['miq_service_id'] = service.id
```
mkanoor: @lfu Can you validate that this has already been implemented in this PR
https://github.com/ManageIQ/manageiq/pull/18057

And is shown here
https://github.com/ManageIQ/manageiq/blob/f2bf17d2a96d01b5ed0a6c8e5870e00307c9cc7b/app/models/service_ansible_tower.rb#L77

Instead of it being id it would be a href slug allowing you to easily access the REST API.

mkanoor:
@h4wkmoon This has been aded recently by @lfu in the core model and it would be available all the time for services as extra variable, hence we won't need this change.
Please let us know if it satisfies your requirement and close this PR if it does

h4wkmoon:
As long as ansible tower receives something to identify the service in cloudforms, I'm happy. :-D

content/automate/ManageIQ/AutomationManagement/AnsibleTower/Service/Provisioning/StateMachines/Provision.class/__methods__/preprovision.rb
h4wkmoon: Perfect. Thanks a lot.

PR #493, opened 2019-01-07
When all child retirement tasks are finished, service retirement is done.

lfu:
Changes to check_service_retired method, so Ansible Tower Service can be retired.

https://bugzilla.redhat.com/show_bug.cgi?id=1650437

@miq-bot add_label bug, blocker, hammer/yes
@miq-bot assign @gmcculloug
@miq-bot add_reviewer @tinaafitz

PR #495, opened 2019-01-09
[WIP] Send into retry if not all child tasks are retired

d-m-u:
https://github.com/ManageIQ/manageiq-content/pull/493, which fixes the service bundle retirement order (the top, root-level service should be retired last) got in yesterday but if the tasks of child services of a service bundle aren't finished retiring, we should be doing a retry. 493 handled the case where all tasks are finished only.  

This should be tested, but there's an existing spec at https://github.com/lfu/manageiq-content/blob/2c4db88147ed9f2207b62efdbf34ed39d22ab542/spec/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/check_service_retired_spec.rb#L17 which I think is passing and probably shouldn't be.

### DIFF
--- a/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/check_service_retired.rb
+++ b/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/check_service_retired.rb
@@ -26,6 +26,9 @@ module ManageIQ
                 if task.miq_request_tasks.all? { |t| t.state == 'finished' }
                   result = 'ok'
                   @handle.log('info', "Child tasks finished.")
+                else
+                  result = 'retry'
+                  @handle.log('info', "Not all child tasks finished.")
                 end
 
                 case result


d-m-u:
@miq-bot add_reviewer @tinaafitz

d-m-u:
@miq-bot add_label bug
I don't have a bug yet though... will update when it's created

tinaafitz:
@d-m-u Looks good. Could you update the spec test?

d-m-u:
@miq-bot add_reviewer @mkanoor

mkanoor:
@d-m-u 
There is already a spec to check the retry how is that working.
https://github.com/ManageIQ/manageiq-content/blob/fcc041546fe82875fe3c787ebc64060d557addbe/spec/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/check_service_retired_spec.rb#L22

Is this function returning retry
https://github.com/ManageIQ/manageiq-content/blob/fcc041546fe82875fe3c787ebc64060d557addbe/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/check_service_retired.rb#L22

If it did it should have worked

d-m-u:
@mkanoor I know, did you read the PR description? I asked Tina to talk to Lucy about it because that test on master should be failing.

d-m-u:
Closed because Lucy's PR is covering this case just fine.

PR #503, opened 2019-01-29
fixing travis: do not create Tag explicitly

yrudman:
`Classification/Tag` designed in a way that `Tag` should not be created explicitly.
It is created (using `:name`) when Classification created


gmcculloug:
@yrudman Do you know what changed caused the tests to break?  Important to know so we can determine if it needs to be back-ported.  Thanks.

yrudman:
this  one, we are still not sure how to proceed
may be the best cause of action just roll-back my changes in 
https://github.com/ManageIQ/manageiq/pull/18378 and work on more save fix

yrudman:
I am rolling back ManageIQ/manageiq#18378

yrudman:
@gmcculloug i reverted braking changes . - https://github.com/ManageIQ/manageiq/pull/18408

gmcculloug:
Ok, do you want to leave this PR open or should it be closed?

yrudman:
I think that test set-up is not correct anyway and this PR legitimate, 
but I do not understand why it is failing ...

gmcculloug:
Thanks for the PR @yrudman.  The test failure was unrelated.  👍

PR #506, opened 2019-02-06
Add specs for storage_hosts that are inaccessible

d-m-u:
Specs for https://github.com/ManageIQ/manageiq-automation_engine/pull/292. 

Since we now have a "... property to the host storage mount information about if a
particular datastore is accessible to the given host or not..." (https://github.com/ManageIQ/manageiq-schema/pull/329) and since we're exposing ```writable_storages```  to run ```writable_accessible_storages``` per https://bugzilla.redhat.com/show_bug.cgi?id=1668020#c26 we should test vmware with storages set to inaccessible.


d-m-u:
@miq-bot assign @gmcculloug 
@miq-bot add_label bug
@miq-bot add_label hammer/no

PR #524, opened 2019-04-11
Removing dependecy on the spec helper file from the main repo

### LINKED ISSUE #6: Test files should not invoke app's spec_helper
Note for us for later, as it's not super important at the moment: We require the main application's spec_helper in `.rspec`. The tests here and on every other cut repo should not require that, as it is a coupling that could cause really weird behavior.

On the plus side, it introduces a way to be able to specify RSpec options from a single place (the main app). If that's something that could be useful, we should store that in an explicit block that is obviously intended, because the coupling here is not obvious from either end. 

cc/ @bdunne

pkomanek:
Removing the requirement for the `spec_helper` file from the manageiq repository.

Fixes #6

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,37 +1,43 @@
 require 'simplecov'
 SimpleCov.start
 
-require 'manageiq-content'
+if ENV["TRAVIS"] || ENV['CI']
+  require 'coveralls'
+  Coveralls.wear!('rails') { add_filter("/spec/") }
+end
 
-# Reset only the ManageIQ automate domain when testing.
-ENV["AUTOMATE_DOMAINS"] = "ManageIQ"
+ENV["RAILS_ENV"] ||= 'test'
+ENV["AUTOMATE_DOMAINS"] = "ManageIQ" # Reset only the ManageIQ automate domain when testing.
+
+require File.expand_path('manageiq/config/environment', __dir__)
+require 'rspec/rails'
 
 def require_domain_file
-  spec_name = caller_locations.first.path
+  spec_name = caller_locations(1..1).first.path
   file_name = spec_name.sub("spec/", "").sub("_spec.rb", ".rb")
-
   AutomateClassDefinitionHook.require_with_hook(file_name)
 end
 
-def domain_file
-  caller_locations(1..1).first.path.sub("spec/", "").sub("_spec.rb", ".rb")
-end
-
+# Requires supporting ruby files
+Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
 Dir[ManageIQ::AutomationEngine::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
 Dir[ManageIQ::Content::Engine.root.join("spec/support/**/*.rb")].each { |f| require f }
 
 RSpec.configure do |config|
   config.include Spec::Support::AutomationHelper
+  config.include Spec::Support::RakeTaskExampleGroup, :type => :rake_task
 
-  config.around(:example) do |ex|
-    begin
-      ex.run
-    rescue SystemExit => e
-      STDERR.puts
-      STDERR.puts "Kernel.exit called from:"
-      STDERR.puts e.backtrace
-      exit 1
-    end
+  config.fixture_path = "#{::Rails.root}/spec/fixtures"
+  config.use_transactional_fixtures = true
+  config.use_instantiated_fixtures  = false
+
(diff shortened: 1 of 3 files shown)


d-m-u:
I'm getting a bunch of 
NameError:
  undefined local variable or method `require_domain_file' for main:Object
  Did you mean?  require_relative
# ./spec/content/automate/ManageIQ/Transformation/Common.class/__methods__/vmtransform_spec.rb:1:in `<top (required)>'

pkomanek:
@d-m-u

d-m-u:
is there a reason it's still WIP?

pkomanek:
@d-m-u are you still getting these errors?All of the test passed fine on my side.

d-m-u:
No, it looks fine now, which is why I asked if there's a reason it's still marked as in progress

gmcculloug:
@bdunne Can you help review this?

spec/spec_helper.rb
```diff
@@ -1,31 +1,70 @@
...
-# Reset only the ManageIQ automate domain when testing.
-ENV["AUTOMATE_DOMAINS"] = "ManageIQ"
+if ENV["TRAVIS"] || ENV['CI']
+  require 'coveralls'
+  Coveralls.wear!('rails') { add_filter("/spec/") }
+end
 
 def require_domain_file
-  spec_name = caller_locations.first.path
+  spec_name = caller_locations(1..1).first.path
```
bdunne: It looks like this method can use the `domain_file` method below and reduce the duplication

spec/spec_helper.rb
```diff
@@ -1,31 +1,70 @@
...
   config.include Spec::Support::AutomationHelper
+  config.include Spec::Support::RakeTaskExampleGroup, :type => :rake_task
+
+  config.fixture_path = "#{::Rails.root}/spec/fixtures"
+  config.use_transactional_fixtures = true
+  config.use_instantiated_fixtures  = false
+
+  # From rspec-rails, infer what helpers to mix in, such as `get` and
+  # `post` methods in spec/controllers, without specifying type
+  config.infer_spec_type_from_file_location!
```
bdunne: I don't think we'll ever have controllers in this repo.  Is this needed?

spec/spec_helper.rb
```diff
@@ -1,31 +1,70 @@
+ENV["RAILS_ENV"] ||= 'test'
+ENV["AUTOMATE_DOMAINS"] = "ManageIQ" # Reset only the ManageIQ automate domain when testing.
+
+require File.expand_path("../manageiq/config/environment", __FILE__)
+require 'rspec/rails'
 require 'simplecov'
-SimpleCov.start
 
-require 'manageiq-content'
+SimpleCov.start
```
bdunne: SimpleCov and Coveralls should be required and started before anything is required.

spec/spec_helper.rb
```diff
@@ -1,31 +1,70 @@
...
+
+  config.mock_with :rspec do |c|
+    c.allow_message_expectations_on_nil = false
+    c.syntax = :expect
+  end
 
-  config.around(:example) do |ex|
+  config.before(:suite) do
+    if ENV["TRAVIS"] && ENV["TEST_SUITE"] == "vmdb"
+      require Rails.root.join('spec', 'coverage_helper.rb')
```
bdunne: I don't think we want to run this here.  It helps verify test coverage on the core rails app.

bdunne: Also, `ENV["TEST_SUITE"]` is never set in this repo.

pkomanek:
@bdunne fixed. I left the `require_domain_file` method and removed `domain_file` method since the 2nd method is not called in this repo.

spec/spec_helper.rb
```diff
@@ -1,31 +1,59 @@
 require 'simplecov'
 SimpleCov.start
 
-require 'manageiq-content'
+if ENV["TRAVIS"] || ENV['CI']
+  require 'coveralls'
+  Coveralls.wear!('rails') { add_filter("/spec/") }
```
bdunne: Is coveralls configured for this repo?

pkomanek: It works with `CI` variable and I think it is also for travis to generate coverage report on websites. Only thing I found in this repository about the `coveralls` gem are these 2 lines inside the `spec_helper` file.

gmcculloug: @bdunne Yes it is, you can see the report as the first comment in the PR: https://github.com/ManageIQ/manageiq-content/pull/524#issuecomment-482217254

PR #540, opened 2019-05-31
ignore .idea directory

branic:
Ignore the .idea directory used by RubyMine. This is already in the .gitignore for the manageiq repo (and probably others).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 /spec/manageiq
 /spec/reports/
 /tmp/
+/.idea/


Fryguy:
You can do this with your global .gitignore... we tend to avoid doing developer specific and OS-specific .gitignore things (it's in manageiq because of old legacy reasons, but I would remove it from there too if I could find a way to avoid upsetting the apple cart)

See http://egorsmirnov.me/2015/05/04/global-gitignore-file.html

As an example - my global .gitignore has the following:

```text
*.sublime-project
*.sublime-workspace
.DS_Store
nbproject/
Thumbs.db
DO_NOT_COMMIT/
.tags
.rubocop-http*
.lockfiles/

coverage/
bower_components/
node_modules/
```

PR #541, opened 2019-05-31
Support handling of special passwords like provision dialogs

branic:
Lifecycle Provisioning has four password fields that end up in the options hash of the provision request (https://github.com/ManageIQ/manageiq-ui-classic/blob/c8a72415ca7b15602745015c6a7e808c7535c3b6/app/views/miq_request/_prov_field.html.haml#L97). This allows the OOTB customization templates to be used effectively.

Before this change service provisioning did not add the encrypted password to the expected fields. (Note: this is the case when the service dialog text box field has the **Protected** attribute enabled). This has an effect of causing the OOTB customization templates to not function when they are referencing the password fields. An example of this is using the **Basic root pass template** which without this change will cause the root password to be set to an empty string.

This PR add handling for these four special password fields that can be passed in from a Service Dialog.

Options Hash Before:

```
{
 ...
 :sysprep_admin_password=>nil,
 ...
 :sysprep_domain_password=>nil,
 ...
 :root_password=>nil,
 ...
 :"password::root_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_root_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_domain_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_domain_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_admin_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_admin_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::user_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_user_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 ...
}
```

Options Hash After:

```
{
 ...
 :sysprep_admin_password=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 ...
 :sysprep_domain_password=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 ...
 :root_password=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 ...
 :"password::root_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_root_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :sysprep_password=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_domain_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_domain_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::sysprep_admin_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_sysprep_admin_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::user_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 :"password::dialog_user_password"=>"v2:{Fp1+1YhW5ssIaA+a7P7v1w==}",
 ...
}
```

### DIFF
--- a/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
+++ b/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
@@ -46,6 +46,14 @@ def add_password_value(sequence, option_key, value, options_hash)
   prefixed_option_key = 'password::dialog_' + option_key
   add_hash_value(sequence, stripped_option_key.to_sym, value, options_hash)
   add_hash_value(sequence, prefixed_option_key.to_sym, value, options_hash)
+
+  if special_password?(option_key.to_sym)
+    add_hash_value(sequence, option_key.to_sym, value, options_hash)
+  end
+end
+
+def special_password?(option_key)
+  [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].include?(option_key) ? true : false
 end
 
 def option_hash_value(dialog_key, dialog_value, options_hash)
--- a/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
+++ b/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
@@ -1,3 +1,4 @@
+SPECIAL_PASSWORD_FIELDS = [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].freeze
 
 def vmdb_object_from_array_entry(entry)
   model, id = entry.split("::")
@@ -53,7 +54,7 @@ def add_password_value(sequence, option_key, value, options_hash)
 end
 
 def special_password?(option_key)
-  [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].include?(option_key) ? true : false
+  SPECIAL_PASSWORD_FIELDS.include?(option_key) ? true : false
 end
 
 def option_hash_value(dialog_key, dialog_value, options_hash)
--- a/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
+++ b/content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
@@ -1,4 +1,4 @@
-SPECIAL_PASSWORD_FIELDS = [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].freeze
+PROVISION_DIALOG_PASSWORD_FIELDS = [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].freeze
 
 def vmdb_object_from_array_entry(entry)
   model, id = entry.split("::")
@@ -48,13 +48,13 @@ def add_password_value(sequence, option_key, value, options_hash)
   add_hash_value(sequence, stripped_option_key.to_sym, value, options_hash)
   add_hash_value(sequence, prefixed_option_key.to_sym, value, options_hash)
 
-  if special_password?(option_key.to_sym)
+  if provision_dialog_password?(option_key.to_sym)
     add_hash_value(sequence, option_key.to_sym, value, options_hash)
   end
 end
 
-def special_password?(option_key)
-  SPECIAL_PASSWORD_FIELDS.include?(option_key) ? true : false
+def provision_dialog_password?(option_key)
+  PROVISION_DIALOG_PASSWORD_FIELDS.include?(option_key) ? true : false
 end
 
 def option_hash_value(dialog_key, dialog_value, options_hash)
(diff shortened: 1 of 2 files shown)


branic:
@mkanoor @gmcculloug This is the PR for the Customization Template issue we were discussing a couple of weeks ago on Gitter. 

I elected to modify dialog_parser instead of CatalogItemInitialization because it is processed first in the state machine. (and it made more sense to me to have it in dialog_parser.)

branic:
I'm not sure why travis or coveralls failed. Any pointers to fix would be appreciated.

It looks to me like the failures are in parts of the code that I didn't touch.

content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
```diff
@@ -46,6 +46,14 @@ def add_password_value(sequence, option_key, value, options_hash)
...
   add_hash_value(sequence, stripped_option_key.to_sym, value, options_hash)
   add_hash_value(sequence, prefixed_option_key.to_sym, value, options_hash)
+
+  if special_password?(option_key.to_sym)
+    add_hash_value(sequence, option_key.to_sym, value, options_hash)
+  end
+end
+
+def special_password?(option_key)
+  [:root_password, :sysprep_password, :sysprep_domain_password, :sysprep_admin_password].include?(option_key) ? true : false
```
gmcculloug: I would love to find a way to use the workflow class method `encrypted_options_fields` so we do not have to duplicate the field names here, but there is a challenge as I do not think we have a direct connection between the service we are processing and the workflow class.  The method is subclassed and exposes different names per sub-class which is the other issue.

See https://github.com/ManageIQ/manageiq/blob/33d994ae207105a508d9e214d39d9ec66ac066dd/app/models/miq_provision_workflow.rb#L34

mkanoor: @branic can you make this as a constant and freeze at the top of the file and then use it here.

content/automate/ManageIQ/Service/Provisioning/StateMachines/Methods.class/__methods__/dialog_parser.rb
```diff
@@ -46,6 +47,14 @@ def add_password_value(sequence, option_key, value, options_hash)
   prefixed_option_key = 'password::dialog_' + option_key
   add_hash_value(sequence, stripped_option_key.to_sym, value, options_hash)
   add_hash_value(sequence, prefixed_option_key.to_sym, value, options_hash)
+
+  if special_password?(option_key.to_sym)
+    add_hash_value(sequence, option_key.to_sym, value, options_hash)
+  end
+end
+
+def special_password?(option_key)
```
gmcculloug: I would suggest we avoid the generic term `special`.  Seem reasonable to replace `special` with `provision_dialog` for the method and constant to give the logic more context.

branic:
@gmcculloug are there any other changes you'd like to see or can this be merged?

gmcculloug:
The code is fine, but would prefer we add something to the existing spec.  I think it could be added here: https://github.com/ManageIQ/manageiq-content/blob/master/spec/automation/unit/method_validation/dialog_parser_spec.rb#L97-L105

kbrock:
@branic Where do these values get stored in the database?
Do we know if `fix_auth` is able to detect these passwords and converts them correctly?

I'm guessing it is in miq_request

gmcculloug:
@kbrock This change sets password values into the options hash of the miq_request_task.  So fix_auth should cover it.

ganeshhubale:
@tinaafitz @gmcculloug @branic 
Please provide test steps or scenario to test this feature. 

Thanks, 
Ganesh

branic:
@ganeshhubale 

This can be tested by:

1. create a Service Dialog with protected text boxes that have the name set as `root_password`, `sysprep_password`, `sysprep_domain_password`, `sysprep_admin_password`.
2. Use the created service dialog in a Service Catalog Item
3. Order the Service Catalog Item
4. The options hash of the request will have the encrypted string for the text as values in the `:root_password`, `:sysprep_password`, `:sysprep_domain_password`, `:sysprep_admin_password` hash keys.

ganeshhubale:
Thank you so much @branic 
I tested with your steps and It is working as expected.

PR #547, opened 2019-06-18
Bug 1721417: ManageIQ | Infrastructure | Retire of retired VM - finis…

vasekbo:
In case of VM has been retired earlier and method /ManageIQ/Infrastructure/VM/Retirement/StateMachines/Methods/finish_retirement end up with and Error messages into automation.log that VM is already retired

### DIFF
--- a/content/automate/ManageIQ/Infrastructure/VM/Retirement/StateMachines/Methods.class/__methods__/finish_retirement.rb
+++ b/content/automate/ManageIQ/Infrastructure/VM/Retirement/StateMachines/Methods.class/__methods__/finish_retirement.rb
@@ -4,6 +4,6 @@
 
 vm = $evm.root['vm']
 if vm && !vm.retired?
- vm.finish_retirement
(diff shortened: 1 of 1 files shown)


d-m-u:
I'm curious about the use case for this since if the VM's already been retired I feel like the situation in which we would be coming back into this method is because the user called retirement twice for the same object. I don't think this is necessary.

lfu:
@vasekbo Please squash your PR into one commit and add some test cases if possible.

tinaafitz:
Hi @vasekbo Thank you for creating a PR.  As @d-m-u mentioned https://github.com/ManageIQ/manageiq-content/pull/547#issuecomment-503073408, the only way the VM would already be marked as retired is if retirement was called multiple times for the same VM.

mfeifer:
@tinaafitz

tinaafitz:
Thanks again for contributing @vasekbo     
Closing the PR because the VM should be available under normal circumstances.

PR #559, opened 2019-08-08
[Transformation][AnsiblePlaybook] Send service_template credential id as part of service dialog

fdupont-redhat:
With Ivanchuk, we identified that the method Transformation/Ansible/LaunchPlaybookAsAService doesn't generate the right service dialog when ordering the Ansible Playbook service template. The initial behavior is to not pass the `credential` field, leading to a nil entry. We discovered that it's now translated into an empty string. Then the credential is not passed to Ansible Runner and the playbook fails to connect to the hosts.

This PR changes the method to include the credential id stored in service template options in the dialog fields. This ensures that the correct credential is picked and passed to Ansible Runner. The behavior has been tested on a live environment and the password is visible in `/tmp/ansible-runner*/env/passwords` file, as expected.

Fixes RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1739247
Depends on: https://github.com/ManageIQ/manageiq-automation_engine/pull/348

### DIFF
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
@@ -27,7 +27,10 @@ module ManageIQ
             return if service_template.nil?
             target_host = target_host(task, transformation_hook)
             return if target_host.blank?
-            service_dialog_options = { :hosts => target_host.ipaddresses.first }
+            service_dialog_options = {
+              :credential => service_template.options[:config_info][:provision][:credential_id],
+              :hosts => target_host.ipaddresses.first
+            }
             service_request = @handle.execute(:create_service_provision_request, service_template, service_dialog_options)
             task.set_option("#{transformation_hook}_ansible_playbook_service_request_id".to_sym, service_request.id)
           rescue => e
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
@@ -29,7 +29,7 @@ module ManageIQ
             return if target_host.blank?
             service_dialog_options = {
               :credential => service_template.options[:config_info][:provision][:credential_id],
-              :hosts => target_host.ipaddresses.first
+              :hosts      => target_host.ipaddresses.first
             }
             service_request = @handle.execute(:create_service_provision_request, service_template, service_dialog_options)
             task.set_option("#{transformation_hook}_ansible_playbook_service_request_id".to_sym, service_request.id)
--- a/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
+++ b/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
@@ -28,7 +28,7 @@ module ManageIQ
             target_host = target_host(task, transformation_hook)
             return if target_host.blank?
             service_dialog_options = {
-              :credential => service_template.options[:config_info][:provision][:credential_id],
+              :credential => service_template.config_info[:provision][:credential_id],
               :hosts      => target_host.ipaddresses.first
             }
             service_request = @handle.execute(:create_service_provision_request, service_template, service_dialog_options)
--- a/spec/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice_spec.rb
+++ b/spec/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice_spec.rb
@@ -42,7 +42,7 @@ describe ManageIQ::Automate::Transformation::Ansible::LaunchPlaybookAsAService d
           :credential_id => credential.id
         }
       }
-     }
+    }
   end
 
   let(:service_template_ansible_playbook_pre) { ServiceTemplateAnsiblePlaybook.create_catalog_item(ansible_playbook_pre_catalog_item_options, user) }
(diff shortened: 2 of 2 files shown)


fdupont-redhat:
@miq-bot add-label blocker, transformation, bug, ivanchuk/yes
@miq-bot add-reviewer @tinaafitz

content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice.rb
```diff
@@ -27,7 +27,10 @@ def main
             return if service_template.nil?
             target_host = target_host(task, transformation_hook)
             return if target_host.blank?
-            service_dialog_options = { :hosts => target_host.ipaddresses.first }
+            service_dialog_options = {
+              :credential => service_template.options[:config_info][:provision][:credential_id],
```
gmcculloug: The `[:config_info]` hash of `service_template` has an accessor and we should use that.

Use either
`service_template.config_info.dig(:provision, :credential_id)`  (If you expect any of the hash levels to be nil)
or
`service_template.config_info[:provision][:credential_id]`

service_template.options[:config_info][:provision][:credential_id]`

fdupont-redhat: Done. Thanks.

gmcculloug:
Can you add a test for this?  Otherwise it looks good, thanks.

spec/content/automate/ManageIQ/Transformation/Ansible.class/__methods__/launchplaybookasaservice_spec.rb
```diff
@@ -0,0 +1,135 @@
+require_domain_file
+
+RSpec::Matchers.define :same_class_and_id do |obj|
+  match { |actual| actual.class == obj.class && actual.id == obj.id }
```
gmcculloug: @fdupont-redhat I have been looking into this more over the weekend and think we should override the `==` method on the [class MiqAeServiceModelBase](https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb#L4) to use the same logic as this matcher then the test would just need to use `eq`.

```ruby
    def ==(other)
      self.id == other.id && self.class == other.class
    end
```

See also [BasicObject#==](https://ruby-doc.org/core-2.5.0/BasicObject.html#method-i-3D-3D)
> Equality — At the Object level, == returns true only if obj and other are the same object. Typically, this method is overridden in descendant classes to provide class-specific meaning.

cc @mkanoor

fdupont-redhat: I proposed https://github.com/ManageIQ/manageiq-automation_engine/pull/348

tinaafitz:
@lfu Please review

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 70c5f459f954f96277920fd073ea43901b303f80
Author: tina <tfitzger@redhat.com>
Date:   Tue Aug 13 13:18:34 2019 -0400

    Merge pull request #559 from fdupont-redhat/v2v_fix_launch_ansible_playbook_credentials
    
    [Transformation][AnsiblePlaybook] Send service_template credential id as part of service dialog
    
    (cherry picked from commit a2fb8c8192327d7a60697fc829b28777f9aa7ca8)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1739247
```

fdupont-redhat:
@miq-bot add-label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 07493fc5b64c55cbe54ef91c1b5c12d21499ced4
Author: tina <tfitzger@redhat.com>
Date:   Tue Aug 13 13:18:34 2019 -0400

    Merge pull request #559 from fdupont-redhat/v2v_fix_launch_ansible_playbook_credentials
    
    [Transformation][AnsiblePlaybook] Send service_template credential id as part of service dialog
    
    (cherry picked from commit a2fb8c8192327d7a60697fc829b28777f9aa7ca8)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1772641
```

NickLaMuro:
@fdupont-redhat I think you might want to double check if this should be in `hammer`.  I am looking into test failures related to the spec added in this PR.

It is failing for a different reason, but you mention `Ansible Runner` in the PR description, and that does **NOT** exist on the `hammer` branch for running playbooks as that was just introduced in the `i-release`.  I am unfamiliar with the specifics of this code, but I would double check that this backport is doing what it should be on that branch, and possibly determine if a different fix is required for `hammer`/5.10

NickLaMuro:
@fdupont-redhat Also, found out what was specs to fail on `hammer` for this repo:

@d-m-u added the ability for class level associations to be automatically exposed in this PR:

https://github.com/ManageIQ/manageiq-automation_engine/pull/308

As it is labeled `hammer/no`, this is why the `config_info` is not available to the model.  You probably can `expose` that method manually here:

https://github.com/ManageIQ/manageiq-automation_engine/blob/95728351f279abcab3140bbb0f891d9c5dc58859/lib/miq_automation_engine/service_models/miq_ae_service_service_template_ansible_playbook.rb#L3

I will open up a patch for `hammer` for that as that most likely the quickest way to get it green again... however, I would still check into what I mentioned above as well just in case.

cc @lfu since this is the issue that you had me look into today.

lfu:
@NickLaMuro Thanks for looking at the issue.

All `virtual_has_one` attributes on Hammer branch would not get exposed in to automate.
For example, a service `virtual_has_one` user. But in automate a call to `service.user` would fail.

simaishi:
@fdupont-redhat We have a build scheduled today. Please let me know what's happening with this.

fdupont-redhat:
@simaishi, I think we should unbackport the 2 PRs. I'll create a Hammer specific PR.

simaishi:
@fdupont-redhat we're good with this PR. @lfu confirmed this change is ok in hammer branch and @NickLaMuro created a PR that fixed exposing association.

fdupont-redhat:
Awesome. Thanks @NickLaMuro for the PR.

PR #571, opened 2019-08-30
Allow for nulls in the exists method

syncrou:
If the allow_null argument is true we now allow any result
to return as true. This nullifies the bool() check
So seems bad, but in some cases when an attribute exists
When looking up 'get_object_attribute_names' we want
the value of that attribute even if it is null
In the case of an allow_null then only a KeyError
will return False

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1730311

### DIFF
--- a/content/ansible/roles/manageiq-core.manageiq-automate/library/manageiq_automate.py
+++ b/content/ansible/roles/manageiq-core.manageiq-automate/library/manageiq_automate.py
@@ -107,19 +107,18 @@ class ManageIQAutomate(object):
         return  json.loads(result.read())
 
 
-    def exists(self, path):
+    def exists(self, path, allow_null=False):
         """
-            Validate all passed objects before attempting to set or get values from them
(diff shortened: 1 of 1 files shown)


syncrou:
@miq-bot assign @tinaafitz 

cc - @mkanoor

syncrou:
@miq-bot add_label bug

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 62108090eb401134c86ba9efa28795e5f710a374
Date:   Tue Sep 3 12:00:18 2019 -0400

    Merge pull request #571 from syncrou/allow_nulls_in_exist
    
    Allow for nulls in the exists method
    
    (cherry picked from commit 2c6c747a81a15cb7b9667644dcb4abff91d7a7c1)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1730311
```

PR #584, opened 2019-09-24
Switch to call lookup_by_* methods

astrozzc:
The find_by_* methods defined in manageiq repo has conflicts with
rails dynamic finding and now they have been changed to lookup_by_*.
Switching the callers in this repo to the lookup_by_* methods.

Depend on https://github.com/ManageIQ/manageiq/pull/19313, https://github.com/ManageIQ/manageiq-automation_engine/pull/373

### DIFF
--- a/content/automate/ManageIQ/Infrastructure/VM/Transform/Import.class/__methods__/list_tag_names.rb
+++ b/content/automate/ManageIQ/Infrastructure/VM/Transform/Import.class/__methods__/list_tag_names.rb
@@ -19,7 +19,7 @@ module ManageIQ
                   category_name = @handle.root['dialog_tag_category']
                   if category_name.present?
                     @handle.log(:info, "Selected tag category: #{category_name}")
-                    category = @handle.vmdb(:classification).find_by_name(category_name)
+                    category = @handle.vmdb(:classification).lookup_by_name(category_name)
                     unless category.nil?
                       category.entries.each do |tag|
                         values_hash[tag.name] = tag.description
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Infrastructure/VM/Transform/Import.class/__methods__/list_tag_names.rb
```diff
@@ -19,7 +19,7 @@ def main
                   category_name = @handle.root['dialog_tag_category']
                   if category_name.present?
                     @handle.log(:info, "Selected tag category: #{category_name}")
-                    category = @handle.vmdb(:classification).find_by_name(category_name)
+                    category = @handle.vmdb(:classification).lookup_by_name(category_name)
```
gmcculloug: @mkanoor @tinaafitz The `find_by_*` calls work today because the automate engine allows these specific calls to pass through, otherwise they need to be specified with `expose` or similar technique in the service_models.

See https://github.com/lfu/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_method_service/miq_ae_service_model_base.rb#L27-L30

Should it now include `lookup_by` and the depreciation side will require some thinking.  Not sure it should hold up these PRs.

mkanoor: @gmcculloug This is ok since its our automate methods, we just have to aware of customers scripts and if they use the find_by_* methods.

PR #569, opened 2019-08-28
[V2V] Remove VMTransform and VMCheckTransformed states

fdupont-redhat:
The state machine is being ported to InfraConversionJob.
This PR is the counterpart of ManageIQ/manageiq#19222.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1746592
Depends on ManageIQ/manageiq#19222
Built on #568

PR #585, opened 2019-09-25
[HAMMER] Add retry_interval to VMCheckTransformed

fdupont-redhat:
So far, we believed that the default retry interval in Automate was 1 minute, and experimentation tended to confirm this. Apparently, it's not the case, as we saw retries every ~4 seconds when no retry interval was specified. This leads to having a TTL for VMCheckTransformed state of about 1 hour and 40 minutes, instead of 25 hours. The consequence is that long migrations fail.

This PR adds a explicit retry interval to ensure that the TTL is effectively 25 hours.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1755632

### DIFF
--- a/content/automate/ManageIQ/Transformation/Common.class/__methods__/vmchecktransformed.rb
+++ b/content/automate/ManageIQ/Transformation/Common.class/__methods__/vmchecktransformed.rb
@@ -11,6 +11,7 @@ module ManageIQ
           def set_retry
             @handle.root['ae_result'] = 'retry'
             @handle.root['ae_retry_server_affinity'] = true
+            @handle.root['ae_retry_interval'] = '1.minutes'
             @handle.log(:info, "Disk transformation is not finished. Checking in #{@handle.root['ae_retry_interval']}")
           end
 


fdupont-redhat:
@miq-bot add-label transformation, bug, blocker/yes, hammer/yes
@miq-bot add-reviewer @tinaafitz 
@miq-bot add-reviewer @gmcculloug

tinaafitz:
@mkanoor Please review.

fdupont-redhat:
@tinaafitz I'm also surprised.
I talked about it with @Fryguy and he told me that the retry interval computation was not as easy as it seems, partly based on a default TTL of 100 minutes, that would be consistent with what we observed, but not with previous observations.

@mkanoor could you explain how this is computed ?

tinaafitz:
Hi @fdupont-redhat Is this code change applicable to Ivanchuk as well?

tinaafitz:
Hi @fdupont-redhat There are multiple ongoing conversations about timeouts. 
Automate has 2 ways to calculate retry intervals and timeouts:
1. The usual Automate state machine - max_retries x the retry interval. 
2. For Ansible services, and playbook methods, the execution_ttl specifies the total time(in minutes) the user expects it to take for the playbook method/service to complete.     Based on the example of execution_ttl = 700 and max_retries = 10, the retry interval would = 70 minutes.

We encountered an issue recently where an Ansible service was timing out after 10 minutes. Automate didn't timeout, but the Ansible runner timeout default value of 10 minutes was exceeded which caused the Service provisioning to fail: https://github.com/ManageIQ/manageiq/pull/19279

We discussed this with Jason and concluded that Automate should always pass a timeout value to Ansible runner. (Before it was passing the execution_ttl, but used the 10 minute default value when execution_ttl was not specified) The timeout value will either be the execution_ttl, or the (max_retries x number of retries)

tinaafitz:
Hi @fdupont-redhat Would this change be applicable to Ivanchuk as well?

fdupont-redhat:
@tinaafitz it still applies to ivanchuk-1, but the series of PR I submitted recently removes the whole method. How would you handle that ?

simaishi:
File removed in https://github.com/ManageIQ/manageiq-content/pull/569 in master.

PR #586, opened 2019-09-27
use :automation_manager_ansible_tower instead of :automation_manager

jameswnl:
Corresponding to the refactoring of factories in https://github.com/ManageIQ/manageiq/pull/19330


jameswnl:
@jrafanie @d-m-u check this out

d-m-u:
I don't think this is the right way to do this. Please see the work done in https://github.com/ManageIQ/manageiq/pull/18842 for https://bugzilla.redhat.com/show_bug.cgi?id=1671844. The factory shouldn't be an intermediate class.

d-m-u:
Then how are we addressing the fact the current issue with the validation in the main repo? I'm fine with this if you've a plan for that as well.

d-m-u:
I don't disagree, but https://github.com/ManageIQ/manageiq/pull/19330/files#diff-4fe1bbb18a78426d823d817a71f11ae6R116 is still going to bite us.

jrafanie:
> I don't think this is the right way to do this. Please see the work done in [ManageIQ/manageiq#18842](https://github.com/ManageIQ/manageiq/pull/18842) for https://bugzilla.redhat.com/show_bug.cgi?id=1671844. The factory shouldn't be an intermediate class.

I think we agree.  For this specific set of tests, they seem to be testing tower automation manager so the change in this PR looks right the right thing to do.

We'll need to fix the base factory as referenced above in a different PR as it's not usable right now.

PR #588, opened 2019-10-10
AzureStack events for targeted refresh

mancabizjak:
In this commit we add automate class and instances corresponding to AzureStack provider events. All the events added here trigger targeted refresh.

#### Links
* Related PR in [ManageIQ/manageiq-providers-azure_stack/pull/16](https://github.com/ManageIQ/manageiq-providers-azure_stack/pull/16) in which we implement `event_target_parser` for processing these events.

@miq-bot assign @gmcculloug 

/cc  @Ladas @agrare

### DIFF
--- a/content/automate/ManageIQ/System/Event/EmsEvent/AzureStack.class/__class__.yaml
+++ b/content/automate/ManageIQ/System/Event/EmsEvent/AzureStack.class/__class__.yaml
@@ -1,0 +1,433 @@
+---
+object_type: class
+version: 1.0
+object:
+  attributes:
+    description: 
+    display_name: 
+    name: AzureStack
+    type: 
+    inherits: 
+    visibility: 
+    owner: 
+  schema:
+  - field:
+      aetype: assertion
+      name: guard
+      display_name: 
+      datatype: string
+      priority: 1
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
+      on_entry: 
+      on_exit: 
+      on_error: 
+      max_retries: 
+      max_time: 
+  - field:
+      aetype: attribute
+      name: logical_event
+      display_name: 
+      datatype: string
+      priority: 2
+      owner: 
+      default_value: 
+      substitute: true
+      message: create
+      visibility: 
+      collect: 
+      scope: 
+      description: 
+      condition: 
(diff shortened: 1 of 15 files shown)


gmcculloug:
@mancabizjak Thanks for the contribution.

One question about the path, where is the naming `Azure_Stack` coming from?  I would expect this to be camel case `AzureStack` which would be consistent with the other EMS events classes: https://github.com/ManageIQ/manageiq-content/tree/master/content/automate/ManageIQ/System/Event/EmsEvent

mancabizjak:
@gmcculloug I share your concern about the consistency in the naming of `Azure_Stack` automate class. The underscore case comes from the current implementation of `EventParser` in AzureStack provider, where we use `AZURE_STACK` source (see [event_parser#L9](https://github.com/ManageIQ/manageiq-providers-azure_stack/blob/9222a1af071b2a59eaec325b720a5ce54e801fdb/app/models/manageiq/providers/azure_stack/cloud_manager/event_parser.rb#L9)), which was autogenerated from the provider name. 

Renaming the class without updating `EventParser` in the provider would give us something like this:

```bash
[----] I, [2019-10-11T13:07:25.811650 #27587:2b28c779a5fc]  INFO -- : Following Relationship [miqaedb:/System/Event/EmsEvent/AZURE_STACK/Administrative_Microsoft.Compute_virtualMachines_write_Succeeded#create]
[----] E, [2019-10-11T13:07:25.867143 #27587:2b28c779a5fc] ERROR -- : Class [System/Event/EmsEvent/AZURE_STACK] not found in MiqAeDatastore
[----] I, [2019-10-11T13:07:25.867176 #27587:2b28c779a5fc]  INFO -- : Followed  Relationship [miqaedb:/System/Event/EmsEvent/AZURE_STACK/Administrative_Microsoft.Compute_virtualMachines_write_Succeeded#create]

```

Personally, I would prefer to name the class `AzureStack` as well, and don't mind making the change in the provider (@agrare does that make sense)?

agrare:
Yeah that works for me, we can make it AZURESTACK, that would be consistent with other providers

mancabizjak:
Okay, done - I renamed the automate class to `AzureStack` and made the change in [ManageIQ/manageiq-providers-azure_stack/pull/17](https://github.com/ManageIQ/manageiq-providers-azure_stack/pull/17). I think we're good to go :+1:

PR #590, opened 2019-10-21
Remove call to retire_service_resources from retire_service method

gmcculloug:
The method has been turned into a no-op method so removing the reference.

Continuing the work identified in https://github.com/ManageIQ/manageiq/pull/16933#discussion_r175805070
and started in https://github.com/ManageIQ/manageiq/pull/19413
which lead to https://github.com/ManageIQ/manageiq-automation_engine/pull/379
😓 

Will need to undo the damage I have done by reverting/closing the referenced PRs.

### DIFF
--- a/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/retire_service.rb
+++ b/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/retire_service.rb
@@ -1,5 +1,5 @@
 #
-# Description: This method attempts to retire all of the vms under this top level service
+# Description: Placeholder method for custom automation
 #
 
-$evm.root["service"].retire_service_resources
+# service = $evm.root["service"]


content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/retire_service.rb
```diff
@@ -1,5 +1,5 @@
 #
-# Description: This method attempts to retire all of the vms under this top level service
+# Description: Placeholder method for custom automation before retiring sub-resources
```
tinaafitz: Hi @gmcculloug Thanks for resolving this issue. :-)
Can you remove  "before retiring sub-services" from the description?
I think that part is misleading since retire tasks are created for all of the Services retireable service resources, and are sent to Automate at the same time.

gmcculloug:
@tinaafitz Change pushed.

PR #606, opened 2019-11-06
Refactor AE Content Rejected method and add spec

GregP:
Quota State Machine Rejected method refactoring.

This PR is based on the issue below:
ManageIQ/manageiq#12038

### DIFF
--- a/content/automate/ManageIQ/System/CommonMethods/QuotaStateMachine.class/__methods__/rejected.rb
+++ b/content/automate/ManageIQ/System/CommonMethods/QuotaStateMachine.class/__methods__/rejected.rb
@@ -1,9 +1,36 @@
+# frozen_string_literal: true
+
 #
 # Description: Quota Exceeded rejected method.
 #
+module ManageIQ
+  module Automate
+    module System
+      module CommonMethods
+        module QuotaStateMachine
+          class Rejected
+            def initialize(handle = $evm)
+              @handle = handle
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/System/CommonMethods/QuotaStateMachine.class/__methods__/rejected.rb
```diff
@@ -1,9 +1,39 @@
...
+            end
+
+            private
+
+            def miq_request
+              @handle.root['miq_request']
+            end
+
+            def rejected
+              @handle.log('info', "Request denied because of #{miq_request.message}")
```
gmcculloug: Not sure the `miq_request` method is buying much here, and in fact it is slightly inefficient compared to the previous logic since it is causing the code to do the `@handle.root['miq_request']` lookup every time it is referenced instead of just once.

I would just keep the previous logic unless this is useful for testing, in which the method should cache the result on the first call and return it for other callers.

billfitzgerald0120:
Can you change this :      expect(ae_service).to receive(:create_notification).once 
To 
expect(ae_service).to receive(:create_notification).with(:level => 'error', :message => 'Quota Exceeded: VM Provisioning - Request Created', :subject => svc_provision_request)

billfitzgerald0120:
@GregP Looks good

PR #610, opened 2019-11-11
[WIP] AE Content refactor Get Retirement Entry Point and add spec

GregP:
Refactor Service Get Retirement Entry Point method and add new spec.

This PR is based on the issue below:
ManageIQ/manageiq#12038

### DIFF
--- a/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
+++ b/content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
@@ -1,15 +1,57 @@
-service = $evm.root['service']
-if service.nil?
-  $evm.log('error', 'get_retirement_entrypoint: missing service object')
-  exit MIQ_ABORT
+# frozen_string_literal: true
+
+#
+# Description: Resolves Service retirement entry point if present
+#
+
+module ManageIQ
+  module Automate
+    module Service
+      module Retirement
+        module StateMachines
+          module Methods
+            class GetRetirementEntrypoint
+              def initialize(handle = $evm)
+                @handle = handle
+              end
+
+              def main
+                retirement_entrypoint
+              end
+
+              private
+
+              def retirement_entrypoint
+                if service
+                  @handle.root['retirement_entry_point'] = entry_point
+                  @handle.root['service_action'] = 'Retirement'
+                  @handle.log("info", "retirement_entrypoint: #{entry_point}")
+                end
+              end
+
+              def service
+                svc = @handle.root['service']
+                if svc.nil?
+                  @handle.log('error', 'retirement_entrypoint: missing service object')
+                  exit MIQ_ABORT
+                end
+                svc
+              end
+
+              def entry_point
+                entry_point = service.automate_retirement_entrypoint
+                if entry_point.blank?
+                  entry_point = '/Service/Retirement/StateMachines/ServiceRetirement/Default'
+                  @handle.log("info", "retirement_entrypoint not specified using default: #{entry_point}")
+                end
+                entry_point
+              end
+            end
+          end
+        end
+      end
+    end
+  end
 end
 
-entry_point = service.automate_retirement_entrypoint
-if entry_point.blank?
-  entry_point = '/Service/Retirement/StateMachines/ServiceRetirement/Default'
-  $evm.log("info", "retirement_entrypoint not specified using default: #{entry_point}")
-end
-
-$evm.root['retirement_entry_point'] = entry_point
-$evm.root['service_action'] = 'Retirement'
-$evm.log("info", "retirement_entrypoint: #{entry_point}")
+ManageIQ::Automate::Service::Retirement::StateMachines::Methods::GetRetirementEntrypoint.new.main


content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
```diff
@@ -1,15 +1,49 @@
...
+              end
+
+              private
+
+              def retirement_entrypoint
+                service = @handle.root['service']
+                if service.nil?
+                  @handle.log('error', 'get_retirement_entrypoint: missing service object')
+                  exit MIQ_ABORT
+                end
```
gmcculloug: @GregP Let's move this into a separate method that either returns the `service` object or raises an error.

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
```diff
@@ -1,15 +1,49 @@
...
+
+                entry_point = service.automate_retirement_entrypoint
+                if entry_point.blank?
+                  entry_point = '/Service/Retirement/StateMachines/ServiceRetirement/Default'
+                  @handle.log("info", "retirement_entrypoint not specified using default: #{entry_point}")
+                end
```
gmcculloug: This should also move into a separate method.

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
```diff
@@ -1,15 +1,49 @@
...
+# Description: Resolves Service retirement entry point if present
+#
+
+module ManageIQ
+  module Automate
+    module Service
+      module Retirement
+        module StateMachines
+          module Methods
+            class GetRetirementEntryPoint
```
lfu: Should this be `GetRetirementEntrypoint`?

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
```diff
@@ -1,15 +1,59 @@
...
+              def retirement_entrypoint
+                if service
+                  @handle.root['retirement_entry_point'] = entry_point
+                  @handle.root['service_action'] = 'Retirement'
+                  @handle.log("info", "retirement_entrypoint: #{entry_point}")
+                end
+              end
+
+              def service
+                @service ||= begin
```
gmcculloug: With the change to remove `if service` in `retirement_entrypoint ` there is no need to cache the results since the `service` method is only called once.

You could simplify to:
```ruby
def service
  if @handle.root['service'].nil?
    @handle.log('error', 'retirement_entrypoint: missing service object')
    exit MIQ_ABORT
  end

  @handle.root['service']
end
```

or even a `tap` version like this:
```ruby
def service
  @handle.root['service'].tap do |svc|
    if svc.nil?
      @handle.log('error', 'retirement_entrypoint: missing service object')
      exit MIQ_ABORT
    end
  end
end
```

lfu: Or it can be simplified to:
```
def service
  @handle.root['service'].tap do |svc|
    raise 'retirement_entrypoint: missing service object'  if svc.nil?
  end
end
```

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
```diff
@@ -1,15 +1,59 @@
...
+              end
+
+              def main
+                retirement_entrypoint
+              end
+
+              private
+
+              def retirement_entrypoint
+                if service
```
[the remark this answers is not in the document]
GregP: @gmcculloug We refer to service when resolving entry point, won't it then try to set it once again?

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
GregP: @gmcculloug Done

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
gmcculloug: @lfu Will the `raise` be processed the same as `exit MIQ_ABORT`?  cc @tinaafitz

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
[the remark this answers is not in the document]
gmcculloug: @GregP The `service` method has two possible results
1. A service object is returned
2. An error (or exit) is raised and the method stops processing.

So either way, `entry_point` will call `service` and either succeed or error out.  It is the same result as before but you do not need an explicit check to validate `service`.

content/automate/ManageIQ/Service/Retirement/StateMachines/Methods.class/__methods__/get_retirement_entrypoint.rb
GregP: @gmcculloug @lfu .tap is roughly 50% slower in run time, and we do need to abort when it is nil, no? I am all for refactoring, but we should not be adding to run time when possible.

gmcculloug: @GregP I would not mind see the timing output from those tests as I expected we are talking about a minimal difference.

My suggestion, and in the spirit of refactoring, would be to do this:
```ruby
def service
  svc = @handle.root['service']
  if svc.nil?
    @handle.log('error', 'retirement_entrypoint: missing service object')
    exit MIQ_ABORT
  end

  svc
end
```

PR #628, opened 2020-01-09
Refactor amazon_customizerequest method for VM Provisioning.

pkomanek:
Refactoring the method and creating tests.

This PR is based on the issue below.
ManageIQ/manageiq#12038

@miq-bot assign @tinaafitz

### DIFF
--- a/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
+++ b/content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
@@ -2,99 +2,166 @@
 # Description: This method is used to Customize the Amazon Provisioning Request
 #
 
-# process_customization - mapping instance_types, key pairs, security groups and cloud-init templates
-def process_customization(mapping, prov, template, product, provider )
-  log(:info, "Processing Amazon customizations...", true)
-  case mapping
-  when 0
-    # No mapping
-  when 1
-    ws_values = prov.options.fetch(:ws_values, {})
+module ManageIQ
+  module Automate
+    module Cloud
+      module VM
+        module Provisioning
+          module StateMachines
+            module Methods
+              class AmazonCustomizeRequest
+                def initialize(handle = $evm)
+                  @handle = handle
+                  @mapping = false
+                end
 
-    if prov.get_option(:instance_type).nil? && ws_values.has_key?(:instance_type)
-      provider.flavors.each do |flavor|
-        if flavor.name.downcase == ws_values[:instance_type].downcase
-          prov.set_option(:instance_type, [flavor.id, "#{flavor.name}':'#{flavor.description}"])
-          log(:info, "Provisioning object updated {:instance_type => #{prov.get_option(:instance_type).inspect}}")
+                def main
+                  log(:info, "Provision:<#{prov.id}> Request:<#{prov.miq_provision_request.id}> Type:<#{prov.type}>")
+                  log(:info, "Template: #{template.name} Provider: #{provider.name} Vendor: #{template.vendor} Product: #{product}")
+                  log(:info, "Processing Amazon customizations...", true)
+                  process_customization if @mapping
+                  log(:info, "Processing Amazon customizations...Complete", true)
+                end
+
+                private
+
+                def prov
+                  @prov ||= @handle.root["miq_provision"].tap do |req|
+                    raise 'miq_provision not specified' if req.nil?
+                  end
+                end
+
+                def template
+                  @template ||= prov.vm_template.tap do |req|
+                    raise 'vm_template not specified' if req.nil?
+                  end
+                end
+
+                def provider
+                  @provider ||= template.ext_management_system.tap do |ems|
+                    raise 'ext_management_system not specified' if ems.nil?
+                  end
+                end
+
+                def product
+                  @product ||= template.operating_system&.product_name&.downcase
+                end
+
+                def ws_values
+                  @ws_values ||= prov.options.fetch(:ws_values, {})
+                end
+
+                def process_customization
+                  set_options
+                  set_customization_template
+                end
+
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,163 @@
...
 
-    if prov.get_option(:guest_access_key_pair).nil? && ws_values.has_key?(:guest_access_key_pair)
-      provider.key_pairs.each do |keypair|
-        if keypair.name == ws_values[:guest_access_key_pair]
-          prov.set_option(:guest_access_key_pair, [keypair.id,keypair.name])
-          log(:info, "Provisioning object updated {:guest_access_key_pair => #{prov.get_option(:guest_access_key_pair).inspect}}")
-        end
-      end
-    end
+                def main(mapping: false)
```
lfu: I think it is better keep the main method signature same as other ruby methods.
```
def main
  ...
  mapping = false
  process_customization(mapping)
 ...
end

def process_customization(mapping)
  if mapping
     set_options
     set_customization_template
  end
end
```

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,163 @@
...
+                      :id => prov.get_option(:customization_template_id).inspect
+                    })
+                    log(:info, "Provisioning object updated {:customization_template_script => %{script}}" % {
+                      :script => prov.get_option(:customization_template_script).inspect
+                    })
+                  end
+                end
+
+                def log(level, msg, update_message = false)
+                  @handle.log(level, msg.to_s)
```
lfu: Is `to_s` necessary? How about `@handle.log(level, msg)`?

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,163 @@
...
+                  @handle.root['miq_provision'].message = msg.to_s if @handle.root['miq_provision'] && update_message
```
lfu: Same here for `to_s`.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,166 @@
...
+                end
+
+                def provider
+                  @provider ||= template.ext_management_system.tap do |ems|
+                    raise 'ext_management_system not specified' if ems.nil?
+                  end
+                end
+
+                def product
+                  @product ||= template.operating_system['product_name'].downcase rescue nil
```
gmcculloug: @pkomanek This refactoring, while it did not introduce the `rescue` statements, should be able to easily remove them.

For most the initial object is guaranteed to be valid because the accessor method is raising an error.  Here the `template` method will always return a `template` object or raise an error.  So we just need to worry about `operating_system` and the remaining items being nil.

If we use the `.` notation instead of `['product_name']` to reference the property you can use safe navigation to achieve this:
```ruby
@product ||= template.operating_system&.product_name&.downcase
```

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,166 @@
...
+                def set_customization_template
+                  if prov.get_option(:customization_template_id).present?
+                    log(:info, "Customization Template selected from dialog ID: %{id} Script: %{script}" % {
+                      :id     => prov.get_option(:customization_template_id).inspect,
+                      :script => prov.get_option(:customization_template_script).inspect
+                    })
+                    return
+                  end
+
+                  customization_template_search_by_function       = "#{prov.type}_#{prov.get_tags[:function]}" rescue nil
```
gmcculloug: The `prov` method will raise an error if it cannot return the object, so we can just remove the `rescue nil` at the end.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,166 @@
...
+                  customization_template_search_by_template_name  = template.name
+                  customization_template_search_by_ws_values      = ws_values[:customization_template] rescue nil
```
gmcculloug: The `ws_values` method defaults to an empty hash so we can just remove the rescue nil at the end.

content/automate/ManageIQ/Cloud/VM/Provisioning/StateMachines/Methods.class/__methods__/amazon_customizerequest.rb
```diff
@@ -2,99 +2,166 @@
...
+                  end
+                  if customization_template.blank?
+                    log(:warn, "Failed to find matching Customization Template", true)
+                  else
+                    log(:info, "Found Customization Template ID: %{id} Name: %{name} Description: %{desc}" % {
+                      :id   => customization_template.id,
+                      :name => customization_template.name,
+                      :desc => customization_template.description
+                    })
+                    prov.set_customization_template(customization_template) rescue nil
```
gmcculloug: This is the only one I'm not really sure about and may want to keep in this refactoring and address later.

PR #666, opened 2020-08-13
Remove refresh_new_target in favor of manager_refresh

agrare:
EmsRefresh.refresh_new_target has been deprecated in favor of EmsRefresh.refresh passing InventoryRefresh::Targets.  RHV is the last user of this method and it can be deleted after this.

Depends on:
- [x] https://github.com/ManageIQ/manageiq-providers-ovirt/pull/515


agrare:
@lufu @tinaafitz PTAL

PR #667, opened 2020-09-10
Plugin Generator Updates

Fryguy:
Plugin Generator Updates

### DIFF
--- a/lib/tasks/README.md
+++ b/lib/tasks/README.md
@@ -1,0 +1,20 @@
+Tasks (.rake files) in this directory will be available as public tasks in the main
+ManageIQ app. They can be executed in the plugin gem via the app: namespace
+
+```shell
+bin/rails app:<task>
+```
+
+Since these tasks are public, please namespace them, as in the following example:
+
+```ruby
+namespace "manageiq:content" do
+  desc "Explaining what the task does"
+  task :your_task do
+    # Task goes here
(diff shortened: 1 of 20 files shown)


Fryguy:
As I ran the plugin generator I adjusted the following:

- I moved some of the helper classes that were directly in the spec/spec_helper.rb into spec/support files
- I effectively undid #524, as I don't understand the rationale for why it was done, other than the original issue being from 2016, but much has changed in pluggability since then.  All of the other plugins follow the pattern as is done in this PR.
- I removed the coveralls bit.  We want to move away from coveralls towards codeclimate for coverage details.

Fryguy:
@gtanzillo Please review.

lib/tasks/README.md
```diff
@@ -0,0 +1,20 @@
...
+```ruby
+namespace "manageiq:content" do
+  desc "Explaining what the task does"
+  task :your_task do
+    # Task goes here
+  end
+end
+```
+
+Tasks places in the lib/tasks_private directory will be private to the plugin
```
gtanzillo, line 19: This should read `Tasks placed ...`

Fryguy: I'll have to change that in the generator and blast it out separately.  Let's leave it here for now.  I have a couple of other things to blast out in the next round.

PR #679, opened 2020-11-13
[available_credentials_spec.rb] Use provider factory

NickLaMuro:
Instead of using the `:embedded_automation_manager_ansible` factory directly, use the `:provider_embedded_ansible` one and reference the `.automation_manager` from the provider.

See this commit for information about the `ensure_manager` code that exists on the provider for `EmbeddedAnsible`:

ManageIQ/manageiq@31a2f329b8809b76c1ef8fce572edde9917973de

Links
-----

Caused by: https://github.com/ManageIQ/manageiq/pull/20787


NickLaMuro:
@agrare Hopefully this should fix the spec failures that I caused.

Going to make a follow up PR in `ManageIQ/manageiq` to :v: deprecate :v: the automation manager factory, with instructions on what to use instead.

NickLaMuro:
@miq-bot add_label bug, test

NickLaMuro:
This is the related PR that I just mentioned:  https://github.com/ManageIQ/manageiq/pull/20816

agrare (approved):

PR #681, opened 2020-12-12
remove attr overriding in validate_quota_spec method

d-m-u:
if attrs is an arg, why are we always initializing it in this method?

### DIFF
--- a/spec/automation/unit/method_validation/validate_quota_spec.rb
+++ b/spec/automation/unit/method_validation/validate_quota_spec.rb
@@ -1,7 +1,7 @@
 describe "Quota Validation" do
   include Spec::Support::QuotaHelper
 
-  def run_automate_method(attrs, used, requested)
+  def run_automate_method(used, requested)
     attrs = []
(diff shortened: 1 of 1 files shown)


tinaafitz (approved):

tinaafitz (approved):

spec/automation/unit/method_validation/validate_quota_spec.rb
```diff
@@ -1,7 +1,7 @@
 describe "Quota Validation" do
   include Spec::Support::QuotaHelper
 
-  def run_automate_method(attrs, used, requested)
+  def run_automate_method(_attrs, used, requested)
```
lfu, line 4: Do we need to keep it as an unused argument?

tinaafitz: Thanks Lucy, that's a good point.  @d-m-u Can we remove the arg?

lfu (approved):

PR #684, opened 2021-01-05
Support the retirement of VM which is provisioned by OVF service template

lfu:
Vm provisioned by service template has an association to service. 

https://github.ibm.com/IBMPrivateCloud/CP4MCM/issues/16312


tinaafitz (commented):
Hi @lfu, 

I think the code change might have an issue with VMs that weren't provisioned by us, and aren't tagged with "retire_full".   

Can we discuss?

lfu:
Closed in favor of https://github.com/ManageIQ/manageiq-providers-vmware/pull/684.

PR #702, opened 2022-02-04
PowerVS additional events

jaywcarman:
Required for:
- https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/335

PR #705, opened 2022-02-08
Clean yaml files

jaywcarman:
Ran `bundle exec rake clean_yaml_files` and committed the generated changes.

This is in response to @Fryguy's comment in https://github.com/ManageIQ/manageiq-content/pull/702#issuecomment-1032708821


Fryguy:
Ugh, the classic whitespace problem rears its ugly head...let's see how this PR fares.

Fryguy (changes requested):
Temporarily blocking this PR...I'd prefer if we didn't have to change all these files, but if we have to, then so be it

jaywcarman:
@agrare took care of this in #707

PR #718, opened 2022-08-31
Do not require prov. dialog to set :number_of_vms

jaywcarman:
For some provisioning dialogs it makes sense to assume the number of VMs
is 1. For example, when provisioning a template from an existing VM. The
VmName suffix method can safely treat a nil :number_of_vms the same as
:number_of_vms == 1, eliminating the need to set the value explicitly in
the provider dialog.

Related to:
- ManageIQ/manageiq-providers-ibm_cloud#418

### DIFF
--- a/spec/content/automate/ManageIQ/Cloud/VM/Provisioning/Naming.class/__methods__/vmname_spec.rb
+++ b/spec/content/automate/ManageIQ/Cloud/VM/Provisioning/Naming.class/__methods__/vmname_spec.rb
@@ -44,12 +44,36 @@ describe ManageIQ::Automate::Cloud::VM::Provisioning::Naming::VmName do
       expect(service.object['vmname']).to eq('abcpro$n{3}')
     end
 
-    it "provisions single vm" do
+    it "provisions single vm without name from dialog" do
       provision.update!(:options => {:number_of_vms => 1})
 
       described_class.new(service).main
 
       expect(service.object['vmname']).to eq('abc$n{3}')
+    end
+
+    it "provisions single vm with name from dialog" do
+      provision.update!(:options => {:number_of_vms => 1, :vm_name => "jay"})
+
+      described_class.new(service).main
+
+      expect(service.object['vmname']).to eq('jay')
+    end
+
+    it "provisions single vm without name from dialog when number of vms is nil" do
+      provision.update!(:options => {})
+
+      described_class.new(service).main
+
+      expect(service.object['vmname']).to eq('abc$n{3}')
+    end
+
+    it "provisions single vm with name from dialog when number of vms is nil" do
+      provision.update!(:options => {:vm_name => "jay"})
+
+      described_class.new(service).main
+
+      expect(service.object['vmname']).to eq('jay')
     end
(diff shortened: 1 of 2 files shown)


jrafanie (approved):
LGTM.  Are there tests for this?

jaywcarman:
> LGTM. Are there tests for this?

There's already a ["provisions single vm"](https://github.com/ManageIQ/manageiq-content/blob/master/spec/content/automate/ManageIQ/Cloud/VM/Provisioning/Naming.class/__methods__/vmname_spec.rb#L47). I could add:
- "provisions with number_of_vms not set"
- "provisions with multiple vms"

Any others?

spec/content/automate/ManageIQ/Cloud/VM/Provisioning/Naming.class/__methods__/vmname_spec.rb
```diff
@@ -44,12 +44,36 @@
...
+    it "provisions single vm without name from dialog" do
       provision.update!(:options => {:number_of_vms => 1})
 
       described_class.new(service).main
 
       expect(service.object['vmname']).to eq('abc$n{3}')
     end
+
+    it "provisions single vm with name from dialog" do
+      provision.update!(:options => {:number_of_vms => 1, :vm_name => "jay"})
```
jaywcarman, line 56: Following the vanity precedent set by "drew" :smile:

jrafanie: Can you add a test for the `suffix` method?

jrafanie: Or maybe `derived_name` if `suffix` is supposed to be private?

jrafanie: Actually, I see.. vmname comes from `derived_name`

jaywcarman: The `provisions single vm with name from dialog` isn't actually affected by the fix - I threw it in for completeness while I was in here - so it will pass with/without.

I did verify locally that both new `...when number of vms is nil` test will fail without d67e8c7.

jrafanie: Thanks... I'm very naive in this area so want to make sure the new tests would protect us in the future.

jrafanie (approved):
:shipit:

spec/content/automate/ManageIQ/Cloud/VM/Provisioning/Naming.class/__methods__/vmname_spec.rb
jaywcarman: No problem, thanks for all of your help!

PR #721, opened 2022-10-21
Multiple notifications bug fix

akhilkr128:
Getting Multiple Notifications in my Services Catalog.
1. Multiple notifications after the service provision
Screenshot 2022-10-19 at 12 36 05 AM
2. Two error messages are showing when an error happens between the provision.
Screenshot 2022-10-19 at 12 36 05 AM

Fix
-----
Added a duplication check before creating notifications.

### DIFF
--- a/content/automate/ManageIQ/Service/Provisioning/StateMachines/ServiceProvision_Template.class/__methods__/update_serviceprovision_status.rb
+++ b/content/automate/ManageIQ/Service/Provisioning/StateMachines/ServiceProvision_Template.class/__methods__/update_serviceprovision_status.rb
@@ -24,10 +24,15 @@ module ManageIQ
                 updated_message = update_status_message(prov, @handle.inputs['status'])
 
                 if @handle.root['ae_result'] == "error"
-                  @handle.create_notification(:level   => "error",
-                                              :subject => prov.miq_request,
-                                              :message => "Service Provision Error: #{updated_message}")
-                  @handle.log(:error, "Service Provision Error: #{updated_message}")
+                  check = "Service Provision Error: #{check_message(prov)}"
+                  message = "Service Provision Error: #{updated_message}"
+                  notified = @handle.vmdb('notification').where('options LIKE ?',"%#{check}%")
+                  if notified.count < 1
+                    @handle.create_notification(:level   => "error",
+                                                :subject => prov.miq_request,
+                                                :message => message)
+                    @handle.log(:error, message)
+                  end
                 end
               end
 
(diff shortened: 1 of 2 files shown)


content/automate/ManageIQ/Service/Provisioning/StateMachines/ServiceProvision_Template.class/__methods__/update_serviceprovision_status.rb
```diff
@@ -24,15 +24,24 @@ def main
...
+                                                :message => message)
+                    @handle.log(:error, message)
+                  end
                 end
               end
 
               private
 
+              def check_message(prov)
+                "Server [#{@handle.root['miq_server'].name}] Service [#{prov.destination.name}]"
```
Fryguy, line 42: I think this check will probably not work when internationalization is enabled, because this would be checking English vs whatever language they have.

content/automate/ManageIQ/Service/Provisioning/StateMachines/ServiceProvision_Template.class/__methods__/update_serviceprovision_status.rb
```diff
@@ -24,15 +24,24 @@ def main
                 updated_message = update_status_message(prov, @handle.inputs['status'])
 
                 if @handle.root['ae_result'] == "error"
-                  @handle.create_notification(:level   => "error",
-                                              :subject => prov.miq_request,
-                                              :message => "Service Provision Error: #{updated_message}")
-                  @handle.log(:error, "Service Provision Error: #{updated_message}")
+                  check = "Service Provision Error: #{check_message(prov)}"
+                  message = "Service Provision Error: #{updated_message}"
+                  notified = @handle.vmdb('notification').where('options LIKE ?', "%#{check}%")
```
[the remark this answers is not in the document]
Fryguy: Additionally, this is a full-table-scan on a text column, so this is likely a very expensive query.

Fryguy:
I'm not sure multiple errors is a problem because if some error actually continually occurs, then it should be said to the user.  The weird part is more the multiple successes.

content/automate/ManageIQ/Service/Provisioning/StateMachines/ServiceProvision_Template.class/__methods__/update_serviceprovision_status.rb
akhilkr128: Got you.

PR #744, opened 2024-05-03
Add a simple email ASL workflow

agrare:


Fryguy (approved):

Fryguy:
I'm going to merge for now, but have some organization thoughts before shipping.

I do wonder if we need to namespace this or something - It's more about organizing the ootb.  We may also need a "system" flag so users can't delete them.

content/workflows/email.asl
```diff
@@ -0,0 +1,20 @@
+{
+  "Comment": "An example of using the builtin scheme.",
```
Fryguy, line 2: Actually, let's describe this properly.  Thinking.
```suggestion
  "Comment": "Send an email.",
```

PR #746, opened 2024-05-03
workflow to provision

kbrock:


content/workflows/provision-vm-builtin/provision-vm-builtin.asl
```diff
@@ -0,0 +1,27 @@
...
+  "States": {
+    "PreProvision": {
+      "Type": "Pass",
+      "Next": "Provision"
+    },
+    "Provision": {
+      "Type": "Task",
+      "Resource": "builtin://provision_execute",
+      "Next": "PostProvision",
+      "Retry": [
```
Fryguy, line 13: Retry is interesting.  Do we have retry on the automate side?

Fryguy (approved):
Overall LGTM, just one quick question

content/workflows/provision-vm-builtin/provision-vm-builtin.asl
kbrock: I was trying to mimic the 'return' status code returned from:

- https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/service_models/miq_ae_service_miq_provision_task.rb#L3
- https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_provision_task.rb

That code is read in `status!` and it is tricky to trigger a retry, so I thought returning a failure would be easier and closer to the way that `floe` / states work

Maybe @agrare has a better suggestion for how this should work?

agrare: What happens when `MiqAeServiceMiqProvisionTask` has status retry?  Is it automatically re-queued or is some external action required?

If we just want floe to say "run the provision, wake me when you're done" then I think we don't want an explicit retry case here it sounds like it is more internal.

kbrock: update:

this is currently copied across all builtin methods to keep running

kbrock:
update:
- dropped the retry logic. (retry is just running)

Fryguy (approved):

kbrock:
:green: (fixed by https://github.com/ManageIQ/manageiq/pull/23027 )

PR #761, opened 2025-03-20
Use the galaxy roles for manageiq_vmdb and manageiq_automate

Fryguy:
Since we have roles on galaxy, we should centralize our code into using those instead of having a duplicate internally. This commit requires a change in core to support galaxy-based out-of-the-box roles for content/ansible

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/23391


Fryguy:
@jrafanie  +5 −1,637 😉

jrafanie (commented):
Do we need to do anything to deprecate and release the existing mechanism to make sure this doesn't change behavior until we go through a release cycle?

Fryguy:
I'm not sure how...I can put it in the radjabov release notes.  Not sure about downstream.

jrafanie (approved):
LGTM

PR #762, opened 2025-03-26
Use specific role versions

Fryguy:
@jrafanie Please review

Alternatively, we can use something like:

```yaml
roles:
- name: manageiq.manageiq_vmdb
  src: https://github.com/ManageIQ/ansible-manageiq-vmdb
  version: v1.0
- name: manageiq.manageiq_automate
  src: https://github.com/ManageIQ/ansible-manageiq-automate
  version: v1.0
```

This would allow this repo to stay more flexible and not have to bump the version every time.  Thoughts?

### DIFF
--- a/content/ansible/roles/requirements.yml
+++ b/content/ansible/roles/requirements.yml
@@ -1,3 +1,5 @@
 roles:
 - name: manageiq.manageiq_vmdb
+  version: v1.0.0
 - name: manageiq.manageiq_automate
+  version: v1.0.0


jrafanie (approved):
LGTM

content/ansible/roles/requirements.yml
```diff
@@ -1,3 +1,5 @@
 roles:
 - name: manageiq.manageiq_vmdb
+  version: v1.0.0
 - name: manageiq.manageiq_automate
+  version: v1.0.0
```
jrafanie, line 5: FYI, since we haven't changed this in so long, I'm fine with staying with more explicit versions.  We can loosen if it we find we need to do more releases.

PR #763, opened 2025-04-01
Add Generic Service Reconfigure entry point

putmanoj:
This PR has changes to Service Reconfigure action, will following changes made.

* adds `/Service/Generic/StateMachines/GenericLifecycle/reconfigure` entry point
* modifies particularly, `PreProcess` method, to pass dialog params (changes from user), for reconfiguration process
* adds `/Service/Generic/StateMachines/Utils/util_object` class, with common methods for `service`, `service_action` & `update_task`,  to use in all the steps methods. 
   * The main reason for adding the 'UtilObject' class is 'Reconfigure' action does not have 'service' & 'service_action' attributes, unlike in case of 'Provision' or 'Retirement' actions
   * task name attribute is different for each of the actions.
   * the methods are common methods used across the steps methods
   

Resolves https://jsw.ibm.com/browse/CP4AIOPS-5515


putmanoj:
> @putmanoj I assume this is mostly copied from another entrypoint, but which one (or is this all brand new)? I just wanted to see them side-by-side so I can understand what I'm looking at.

Hi @Fryguy 
I will add all the PRs  in the channel. 
Also, updated the description, for better understanding.

Acutally, probably methods can be removed, we could use default methods, if  ...

- 'ReconfigureStart' (method) :  if we find a way to set 'service_action' ae attribute in ManageIQ core for reconfigure task - _**needs more time to debug, to find out how to set.**_

- 'ReconfigurePreProcess'  (method) -  if we merge to default PreProcess, as I feel this is not a big change https://github.com/ManageIQ/manageiq-content/pull/763/files#diff-7be84ec5751a81c8a14c2ee722a8efbfc5290f017087c4fb5e9d6a67fa11d2c0R78. -

content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/preprocess.rb
```diff
@@ -32,31 +32,31 @@ def main
...
+              def reconfiguration_options
+                task = @handle.root["service_reconfigure_task"]
+                {:dialog => task.options[:dialog] || {}}
+              end
+
+              def update_options
+                if service_action == 'Reconfigure'
+                  reconfiguration_options
+                else
+                  {}
```
agrare, line 59: Note to self: something about this doesn't seem right in the generic lifecycle preprocess, needs investigation (is there a reconfigure override we can stick this in?)

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
```diff
@@ -0,0 +1,60 @@
...
+                end
+              end
+
+              def self.service(handle = $evm)
+                if handle.root["request"] == "service_reconfigure"
+                  task = handle.root["service_reconfigure_task"]
+                  service = task.source unless task.nil?
+                else
+                  service = handle.root["service"]
+                end
```
agrare, line 36: Having this kind of "is it provision or is it reconfigure" logic in a generic utils object also doesn't feel right

putmanoj: Created the utils object (probably could have named it 'common' object), was to hide this _"is it provision or is it reconfigure"_ logic, from the step methods.  

At present, 
- `service` method - **need not have** the _"is it provision or is it reconfigure"_ logic, as the 'service' object gets added with  https://github.com/ManageIQ/manageiq/pull/23405/files#diff-7bcaccf0a5393207d24e32b17c0758232ab7f73c9ae41412c4b25e16b06f5eeaR39 change

- `service_action` method - needs the _"is it provision or is it reconfigure"_ logic, as did not yet find the way to add the 'service_name' attribute.

- `update_task` method - needs the  _"is it provision or is it reconfigure"_ logic, as simply the task names are named differently, for Provision vs Reconfigure vs Retirement.

Also, would say there is bug a in existing update_task method for Retirement,  see https://github.com/putmanoj/manageiq-content/blob/master/content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/execute.rb#L34

```
              def update_task(message)
                @handle.root['service_template_provision_task'].try { |task| task.miq_request.user_message = message }
              end
```
though for Retirement the task_name is 'service_retire_task'


For Provision
Screenshot 2025-04-09 at 5 19 57 PM


For Reconfigure
Screenshot 2025-04-09 at 5 18 33 PM

For Retirement
Screenshot 2025-04-09 at 5 17 49 PM

content/automate/ManageIQ/Service/Generic/StateMachines/GenericLifecycle.class/__methods__/preprocess.rb
putmanoj: Yes, agree, needs investigation,  though for now, did not find anything better than this.

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
```diff
@@ -0,0 +1,63 @@
...
+                            end
+                handle.root[task_name].try do |task|
+                  task.miq_request.user_message = message
+                  task.message = status unless status.nil?
+                end
+              end
+
+              def self.service(handle = $evm)
+                if handle.root["request"] == "service_reconfigure"
+                  task = handle.root["service_reconfigure_task"]
```
agrare, line 35: If we set the service here could we drop this conditional?  Seems weird that we have service for prov and retire but not reconfigure

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
```diff
@@ -0,0 +1,63 @@
...
+              end
+
+              def self.update_task(message, status = nil, handle = $evm)
+                task_name = case handle.root["request"]
+                            when "service_reconfigure"
+                              "service_reconfigure_task"
+                            when "service_retire"
+                              "service_retire_task"
+                            else
+                              "service_template_provision_task"
```
agrare, line 25: What is `handle.root["request"]` here?  Is it `service_template_provision`?  If so could we generically do `#{handle.root["request"]}_task}`?

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
```diff
@@ -0,0 +1,63 @@
...
+                            end
+                handle.root[task_name].try do |task|
+                  task.miq_request.user_message = message
+                  task.message = status unless status.nil?
```
agrare, line 29: Really minor but we tend to prefer the "positive" `if status` over the "negative" `unless status.nil?`

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
agrare: You'll notice I'm trying to kill this whole util_object if we can get away with it :)

putmanoj: Check images attached in https://github.com/ManageIQ/manageiq-content/pull/763#discussion_r2035233031

* Provision 
   * request =>  "clone_to_service"
   * service_action => "Provision"
* Reconfigure
   * request =>  "service_reconfigure"
   * service_action => (NOT AVAILABLE)  - _this is the main reason, need to add the util_object, unless we can fix this in core_
* Retirement
   * request =>  "service_retire"
   * service_action => "Retirement"

putmanoj: Apart from the task name, where patterns are different, between the actions

content/automate/ManageIQ/Service/Generic/StateMachines/Utils.class/__methods__/util_object.rb
```diff
@@ -0,0 +1,64 @@
...
+                end
+              end
+
+              def self.service(handle = $evm)
+                service = handle.root["service"] || service_task(handle)&.source
+                service.tap do |s|
+                  ManageIQ::Automate::System::CommonMethods::Utils::LogObject.log_and_raise("Service not found", handle) if s.nil?
+                end
+              end
```
agrare, line 48: @putmanoj with https://github.com/ManageIQ/manageiq/pull/23405 merged do we still need this ?

PR #791, opened 2026-06-23
Make RemoveFromInventory false by default

putmanoj:
When we retire, don't remove from inventory by default. 

For example, in case of EmbeddedTerraform we need the retired Stack, to fetch stdout from backend.

### DIFF
--- a/content/workflows/retire.asl
+++ b/content/workflows/retire.asl
@@ -13,7 +13,7 @@
       "Parameters": {
         "RemoveFromProvider": true,
         "RemoveFromProviderStorage": true,
-        "RemoveFromInventory": true
+        "RemoveFromInventory": false
       },
       "Next": "PostRetire"
     },


agrare:
By default in automate we remove Cloud and Infra VMs as well as OrchestrationStacks from inventory after retirement:
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Cloud/Orchestration/Retirement/StateMachines/StackRetirement.class/__class__.yaml#L234-L253
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Cloud/Orchestration/Retirement/StateMachines/StackRetirement.class/__class__.yaml#L234-L253
https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Infrastructure/VM/Retirement/StateMachines/VMRetirement.class/default.yaml#L15-L16

I don't think we should just change the default here from true to false.

We should either use Input to let the user override this value or create a separate retire asl for automation.

putmanoj:
@agrare 
This would then need another Issue created, can't get this done with current issue.
I feel  it is better to close current PR,  when resources successfully destroyed, the stdout is not very critical for the user. 
We can get back this, at a later stage, when we plan more UI changes.

